PR #52, opened 2023-04-06
docs: raw REDME thoughts

w3b6x9:


chasers:
Thanks @w3b6x9 will pull in some edits from this.

PR #65, opened 2023-04-10
docs(openapi): match tenant update implementation

ChristianAlexander:
The `PUT /api/tenants/{external_id}` controller method accepts a JSON body containing a partial tenant—excluding `id`, `external_id`, and timestamps—all wrapped in a `tenant` key.

External ID is obtained through a path parameter.

This change updates the definition of the body to match the controller implementation.

This PR updates documentation that lives within application code.

Currently, the documentation does not reflect the implemented behavior.

## What is the new behavior?

Screen Shot 2023-04-10 at 7 31 27 AM

## Additional context


abc3 (approved):
Thanks for the contribution ❤️

PR #66, opened 2023-04-10
feat: remove unused Phoenix generator content

ChristianAlexander:

- Removes unused modules and an unused Phoenix Framework image file.
- Updates `root.html.heex` to stop referencing CSS and JS files that do not exist.

Currently, some content from the original Phoenix generation of the project remains in the project.
These files can be distracting and may be distributed in Docker images / burrito binaries—leading to marginally larger artifacts.

## What is the new behavior?

Some of the vestigial files from the original Phoenix project generation are no longer present in the repository.

## Additional context

- I didn't go so far as to remove `root.html.heex` altogether, in case a user interface is ever desired.
- I noticed that `app.css` and `app.js` were already removed, so I removed dead links to them.


abc3 (approved):
❤️

PR #73, opened 2023-04-21
Implement multiple pools per tenant

abc3:
This pull request introduces the capability to start multiple connection pools for tenants, organized by user entity.

Connection pools are identified using the pair `{tenant, db_user_alias}`.

The `db_user_alias` field serves as an alias for the `db_user` field, allowing for the creation of different types of pools (transaction/session/etc.) with the same Postgres user simultaneously.

A tenant can have many users, but the following combination for a user should be unique:
```elixir
[:db_user_alias, :tenant_external_id, :mode_type]
```

Supavisor retrieves the user name and tenant from the Postgres `username` field, which is divided by `.`. If a tenant has only one user, it is allowed to use just the tenant name in the username field, for example:
```bash
postgresql://some_tenant@localhost:7654/postgres
```

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -12,7 +12,7 @@ defmodule Supavisor.ClientHandler do
   @behaviour :gen_statem
 
   alias Supavisor.DbHandler, as: Db
-  alias Supavisor.{Tenants, Tenants.Tenant, Protocol.Server, Monitoring.Telem, Manager}
+  alias Supavisor.{Tenants, Tenants.User, Protocol.Server, Monitoring.Telem, Manager}
 
   @impl true
   def start_link(ref, _socket, transport, opts) do
@@ -43,6 +43,7 @@ defmodule Supavisor.ClientHandler do
       trans: trans,
       db_pid: nil,
       tenant: nil,
+      user_alias: nil,
       pool: nil,
       manager: nil,
       query_start: nil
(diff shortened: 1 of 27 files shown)


lib/supavisor/client_handler.ex
```diff
@@ -101,12 +99,12 @@ defmodule Supavisor.ClientHandler do
     end
   end
 
-  def handle_event(:internal, :subscribe, _, %{tenant: tenant} = data) do
-    Logger.info("Subscribe to tenant #{tenant}")
+  def handle_event(:internal, :subscribe, _, %{tenant: tenant, user_alias: db_alias} = data) do
+    Logger.info("Subscribe to tenant #{inspect({db_alias, tenant})}")
```
wojtekmach, line 103: nitpick, in many other places we'd log `{tenant, db_alias}` and so being consistent here might help debugging later, i.e. searching for the same thing:

```suggestion
    Logger.info("Subscribe to tenant #{inspect({tenant, db_alias})}")
```

lib/supavisor/tenants/user.ex
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+    user
+    |> cast(attrs, [
+      :db_user_alias,
+      :db_user,
+      :db_password,
+      :pool_size,
+      :mode_type,
+      :is_manager
```
wojtekmach, line 38: double-checking: how important is is_manager field? Can any API user set it? Usually what we might want to do is allow setting this field only on creation but not on update. Or, allow it on update but only under certain circumstances. In those cases it's usually better not to have the field in the cast allowlist but instead set it manually using functions like `put_change` when those extra conditions are met.

wojtekmach (commented):
Looks good to me!

lib/supavisor/client_handler.ex
abc3: Yep, right, thanks!

lib/supavisor/tenants/user.ex
abc3: This field is not currently in use, but we plan to use it for internal operations in the future, such as obtaining the `max_connections` value. Thank you for the relevant thoughts, I'll keep them in mind 🔥

PR #76, opened 2023-04-24
feat: add release workflow for burrito

J0:
## What kind of change does this PR introduce?

Drive by contribution since the workflow seems similar to what we have in Auth. Aims to fix #13 , let me know if anything should be amended

How this was tested:

- [x] Tested via [Act](https://github.com/nektos/act) on a [fork](https://github.com/J0/supavisor/actions/runs/5568478329)

## Additional Notes

The action needs to be granted write permissions to  upload the binaries - Settings > Actions  > Workflow Permissions > Read and Write Permissions

PR #78, opened 2023-04-30
feat: add version check to CI via VERSION file

J0:

PR for semver release workflow derived from the PR in #76  . Based on discussion it sounds like we  want either:

- Logflare has [a VERSION file](https://github.com/Logflare/logflare/blob/staging/VERSION) we do things with.
- Realtime has a [mix version check](https://github.com/supabase/realtime/blob/main/.github/workflows/version_updated.yml) so it reminds us to update the version there too.
- Something else

Left as draft for now, can decide what to do in the coming week.


J0:
Hey @chasers, @abc3,

Just wanted to circle back and check in on this - any preference between:

1. Logflare has [a VERSION file](https://github.com/Logflare/logflare/blob/staging/VERSION) we do things with.

2. Realtime has a [mix version check](https://github.com/supabase/realtime/blob/main/.github/workflows/version_updated.yml) so it reminds us to update the version there too.

3. Other

Was hoping to close this out before I return from OOO on Wed evening PST.  Can also close this for now if more discussion is needed.

Let me know!

chasers:
@J0 let's do the VERSION file like Logflare. And how we pull in the version from that into the mix.exs file.

J0:
Hey @chasers,

Not sure if this is what you had in mind - this imposes a check which requires the dev to manually update the `VERSION` file for the PR similar to [how it is done in realtime](https://github.com/supabase/realtime/blob/main/.github/workflows/version_updated.yml#L24)

Let me know if you had something else in mind and I can adjust accordingly.

chasers:
Thanks so much! LGTM!

chasers (approved):

PR #79, opened 2023-05-02
Implement session mode functionality

abc3:
In this PR, session mode functionality is implemented, with each client connection initiating a direct link to the tenant's database. The primary work is done using poolboy's `max_overflow` parameter. 

For session mode, we start an empty pool and set a specific value for `max_overflow`, which represents the maximum number of workers created if the pool is empty.

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -366,5 +366,5 @@ defmodule Supavisor.ClientHandler do
     nil
   end
 
-  defp handle_db_pid(:ssession, _, db_pid), do: db_pid
+  defp handle_db_pid(:session, _, db_pid), do: db_pid
 end
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -358,7 +358,8 @@ defmodule Supavisor.ClientHandler do
     db_pid
   end
 
-  @spec handle_db_pid(:transaction | :session, pid(), pid()) :: nil | pid
+  @spec handle_db_pid(:transaction, pid(), pid()) :: nil
+  @spec handle_db_pid(:session, pid(), pid()) :: pid()
   defp handle_db_pid(:transaction, pool, db_pid) do
     Process.unlink(db_pid)
     :poolboy.checkin(pool, db_pid)
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -218,18 +218,11 @@ defmodule Supavisor.ClientHandler do
     if ready? do
       Logger.debug("Client is ready")
 
-      new_data =
-        if data.mode == :transaction do
-          Process.unlink(data.db_pid)
-          :poolboy.checkin(data.pool, data.db_pid)
-          %{data | db_pid: nil}
-        else
-          data
-        end
+      db_pid = handle_db_pid(data.mode, data.pool, data.db_pid)
 
       Telem.network_usage(:client, data.socket, data.tenant, data.user_alias)
       Telem.client_query_time(data.query_start, data.tenant, data.user_alias)
-      {:next_state, :idle, new_data, reply}
+      {:next_state, :idle, %{data | db_pid: db_pid}, reply}
     else
       Logger.debug("Client is not ready")
       {:keep_state_and_data, reply}
@@ -364,4 +357,13 @@ defmodule Supavisor.ClientHandler do
     Telem.pool_checkout_time(time, tenant, user_alias)
     db_pid
   end
+
+  @spec handle_db_pid(:transaction | :session, pid(), pid()) :: nil | pid
+  defp handle_db_pid(:transaction, pool, db_pid) do
+    Process.unlink(db_pid)
+    :poolboy.checkin(pool, db_pid)
+    nil
+  end
+
+  defp handle_db_pid(:ssession, _, db_pid), do: db_pid
 end
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -115,16 +115,12 @@ defmodule Supavisor.ClientHandler do
            Supavisor.subscribe_global(node(tenant_sup), self(), tenant, db_alias),
          ps <-
            Manager.get_parameter_status(manager) do
-      db_pid =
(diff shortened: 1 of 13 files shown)


lib/supavisor/tenants/user.ex
```diff
@@ -13,8 +13,9 @@ defmodule Supavisor.Tenants.User do
     field(:db_user, :string)
     field(:db_password, Supavisor.Encrypted.Binary, source: :db_pass_encrypted)
     field(:is_manager, :boolean, default: false)
-    field(:mode_type, Ecto.Enum, values: [:transaction])
+    field(:mode_type, Ecto.Enum, values: [:transaction, :session])
     field(:pool_size, :integer)
+    field(:pool_checkout_timeout, :integer, default: 60000)
```
abc3, line 18:
```suggestion
    field(:pool_checkout_timeout, :integer, default: 60_000)
```

test/integration/proxy_test.exs
```diff
@@ -124,4 +124,18 @@ defmodule Supavisor.Integration.ProxyTest do
...
+
+  test "too many clients in session mode" do
+    db_conf = Application.get_env(:supavisor, Repo)
+
+    url =
+      "postgresql://session.proxy_tenant:#{db_conf[:password]}@#{db_conf[:hostname]}:#{Application.get_env(:supavisor, :proxy_port)}/postgres"
+
+    spawn(fn -> System.cmd("psql", [url], stderr_to_stdout: true) end)
+
+    :timer.sleep(500)
```
wojtekmach, line 136: Right, we need to wait to make sure the client connection is up. Another way of solving this would be to allow hooking into the proxy lifecycle. For example, with Postgrex (thanks to the underlying DBConnection) we can do this:

```elixir
pid = self()

{:ok, _} =
  Postgrex.start_link(..., after_connect: fn _conn ->
    send(pid, :connected)
  end)

assert_receive :connected, 500
```

which is really convenient in tests. Perhaps it's something worth considering.

Although in this particular case it could be a bit difficult to hook into given it seems we're not explicitly starting proxy and instead we rely on the one started in Supavisor's application supervision tree and so the only way to configure it would be using application env which then would affect other tests. So yeah unfortunately it's not trivial but perhaps worth pursuing after all in other PRs.

wojtekmach (commented):
Looks good to me!

lib/supavisor/client_handler.ex
```diff
@@ -309,4 +349,21 @@ defmodule Supavisor.ClientHandler do
...
+  end
+
+  @spec handle_db_pid(:transaction | :session, pid(), pid()) :: nil | pid
+  defp handle_db_pid(:transaction, pool, db_pid) do
+    Process.unlink(db_pid)
+    :poolboy.checkin(pool, db_pid)
+    nil
+  end
+
+  defp handle_db_pid(:ssession, _, db_pid), do: db_pid
```
josevalim, line 368: Typo here `:ssession`. Perhaps this clause is not being covered?

lib/supavisor/client_handler.ex
```diff
@@ -107,9 +115,16 @@ defmodule Supavisor.ClientHandler do
            Supavisor.subscribe_global(node(tenant_sup), self(), tenant, db_alias),
          ps <-
            Manager.get_parameter_status(manager) do
+      db_pid =
+        if data.mode == :session do
+          db_checkout(pool, tenant, db_alias, data.timeout)
```
josevalim, line 120: What if we call it as: 

```suggestion
          db_checkout(data.mode, pool, tenant, db_alias, data.timeout)
```

So all of the session vs transaction logic is encapsulated in helper functions at the end of the module?

josevalim:
Looks good to me. The max_overflow trick is a nice one. In theory, for session mode we wouldn't even need poolboy, but keeping the unified workflow is certainly a benefit.

Just keep in mind you shouldn't really mix transaction mode and session mode in the same pool. I don't think it is done right now, but worth saying. :) That's because we could end-up with a pathological scenario where a pool of size 10 and max_overflow 10 has 10 session clients taking all "size connections" and then all transaction clients using the max_overflow ones (which would require explicit reconnections every time).

lib/supavisor/client_handler.ex
```diff
@@ -309,4 +349,21 @@ defmodule Supavisor.ClientHandler do
...
+
+  @spec db_checkout(pid(), String.t(), String.t(), integer()) :: pid()
+  defp db_checkout(pool, tenant, user_alias, timeout) do
+    {time, db_pid} = :timer.tc(:poolboy, :checkout, [pool, true, timeout])
+    Process.link(db_pid)
+    Telem.pool_checkout_time(time, tenant, user_alias)
+    db_pid
+  end
+
+  @spec handle_db_pid(:transaction | :session, pid(), pid()) :: nil | pid
```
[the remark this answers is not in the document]
abc3: nice!

lib/supavisor/client_handler.ex
```diff
@@ -127,11 +142,14 @@ defmodule Supavisor.ClientHandler do
     :keep_state_and_data
   end
 
-  def handle_event(:info, {:tcp, _, bin}, :idle, data) do
+  def handle_event(:info, {:tcp, _, bin}, :idle, %{mode: :session} = data) do
```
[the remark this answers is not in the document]
abc3: but it is named as data in the [gen_statem specification](https://www.erlang.org/doc/man/gen_statem.html#type-data)?

lib/supavisor/client_handler.ex
abc3, line 369:
```suggestion
  defp handle_db_pid(:session, _, db_pid), do: db_pid
```

test/integration/proxy_test.exs
abc3: Thanks, that is very interesting. I will try it in a separate commit 👍

abc3:
@josevalim I completely agree with your point of view, and maintaining a unified workflow was my goal. Thank you for pointing out the potential issue of mixing transaction and session modes. This is definitely something to keep in mind 👍

abc3:
@334499p by default the session mode works on the `5432` port

PR #82, opened 2023-05-05
Implement Websocket proxy

abc3:


chasers (approved):

josevalim (commented):
LGTM!

The only thing I would add is that Phoenix `socket` normalizes the communication so it supports both longpolling and websockets on top of the same socket.

However, latest Plug and Phoenix come with builtin delegation to connection upgrades and websockets, in case you want to skip the the additional layer of indirection: https://hexdocs.pm/plug/Plug.Conn.html#upgrade_adapter/3

lib/supavisor_web/ws_proxy.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  def handle_in({bin, _opts}, %{socket: socket} = state) do
+    :ok = :gen_tcp.send(socket, bin)
+    {:ok, state}
+  end
+
+  def handle_info({:tcp, _, bin}, %{status: :startup} = state) do
+    if String.ends_with?(bin, Server.ready_for_query()) do
+      acc = filter_pass_pkt(state.acc)
+      :ok = :gen_tcp.send(state.socket, acc)
```
josevalim, line 47: FWIW, this may return `{:error, _}`, but letting it crash is fine (and probably what would be done anyway).

lib/supavisor_web/endpoint.ex
```diff
@@ -10,6 +10,14 @@ defmodule SupavisorWeb.Endpoint do
     signing_salt: "zJOrGxcM"
   ]
 
+  socket "/socket/vercel", SupavisorWeb.WsProxy,
```
chasers, line 13: Hmm on second thought maybe we keep "/vercel" out of this. Are there any other clients which implement the wire protocol over websockets?

lib/supavisor_web/endpoint.ex
```diff
@@ -10,6 +10,14 @@ defmodule SupavisorWeb.Endpoint do
...
+    websocket: [path: "/"],
+    longpoll: false
+
+  socket "/v2", SupavisorWeb.WsProxy,
```
chasers, line 17: Also this we should handle in the API gateway and just route /v2 to /socket.

But it's okay in here for now for testing until we get everything else setup.

abc3:
@josevalim, thanks for the advice about `Plug.Conn.upgrade_adapter/3`. Can you please review my latest commit to check if I did everything right? https://github.com/supabase/supavisor/pull/82/commits/9c987f11219a45bd9e2d24231961473310ef35c5

lib/supavisor_web/router.ex
```diff
@@ -38,6 +38,11 @@ defmodule SupavisorWeb.Router do
     get("/openapi", OpenApiSpex.Plug.RenderSpec, [])
   end
 
+  # websocket pg proxy
+  scope "/v2" do
+    get("/", SupavisorWeb.WsProxy, :index)
```
abc3, line 43:
```suggestion
    get("/", SupavisorWeb.WsProxy, [])
```

josevalim:
Yup, looks great to me. It should indeed be mostly equivalent, just less indirection. :)

PR #95, opened 2023-05-23
Systemd service

abc3:
The current PR introduces a config template and incorporates several environment variables that enable Supavisor to run as a `systemd` service. Additionally, it provides the possibility to use the `Epmd` libcluster strategy.


config/runtime.exs
```diff
@@ -1,39 +1,67 @@
...
+topologies = []
+
+topologies =
+  if System.get_env("DNS_POLL") do
+    dns_poll = [
+      strategy: Cluster.Strategy.DNSPoll,
+      config: [
+        polling_interval: 5_000,
+        query: System.get_env("DNS_POLL"),
+        node_basename: System.get_env("FLY_APP_NAME") || "supavisor"
```
abc3, line 33:
```suggestion
        node_basename: System.get_env("NODE_NAME") || System.get_env("FLY_APP_NAME") || "supavisor"
```

deploy/service/supavisor.service
```diff
@@ -0,0 +1,21 @@
...
+EnvironmentFile=/opt/supavisor/service_vars.ini
+WorkingDirectory=/opt/supavisor
+ExecStart=/opt/supavisor/bin/server start
+ExecStop=/opt/supavisor/bin/server stop
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+TasksMax=infinity
```
abc3, line 21:
```suggestion
TasksMax=infinity

```

deploy/service/supavisor.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=Realtime
```
abc3, line 2:
```suggestion
Description=Supavisor
```

deploy/service/service_vars.ini
```diff
@@ -0,0 +1,16 @@
+MIX_ENV=prod
```
abc3, line 1:
```suggestion
MIX_ENV=""
```

deploy/service/supavisor.service
```diff
@@ -0,0 +1,21 @@
...
+Description=Supavisor
+After=network.target
+Requires=network.target
+StartLimitIntervalSec=5
+
+[Service]
+Type=simple
+Restart=always
+RestartSec=5
+User=root
```
philss, line 11: I never had to create a service file like this before, but I remember PostgreSQL runs with a different user - the `postgres` user, like the docs suggests: https://www.postgresql.org/docs/current/server-start.html -, and I think that's a good practice.

I wonder if you could run with a non-privilege user here. Do you need something that requires `root`?

philss (commented):
I added two minor points, but overall it looks good to me! :+1:

config/runtime.exs
```diff
@@ -1,39 +1,68 @@
...
-        socket_opts: [:inet6]
+config :supavisor, SupavisorWeb.Endpoint,
+  server: true,
+  http: [
+    port: String.to_integer(System.get_env("PORT") || "4000"),
+    transport_options: [
+      max_connections: String.to_integer(System.get_env("MAX_CONNECTIONS") || "1000"),
+      num_acceptors: String.to_integer(System.get_env("NUM_ACCEPTORS") || "100"),
+      socket_opts: [
+        System.get_env("ADDR_TYPE", "inet") |> String.to_atom()
```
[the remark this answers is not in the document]
abc3: good call 👍

deploy/service/supavisor.service
abc3: If I'm not mistaken, only root can listen to ports lower than 1024. I'm considering including a 'stub user' as an example, since we have the option to choose which port Phoenix starts on

abc3, line 11:
```suggestion
User=${SUPAVISOR_USER}
```

PR #101, opened 2023-05-31
Check creds before adding to db and wait timeout for cold start

abc3:
This PR introduces a feature for validating database credentials and fixes a bug in `ClientHandler` which previously did not send parameter status packets during the initial connection if there was no connection to the tenant's database.

The validation process involves selecting the current version of the PostgreSQL connection using the database credentials when they are added via the API. This information is then stored in the `default_parameter_status` column of the tenant record as `%{"server_version": version}`.

The workflow of parameter status packets is as follows:
- During initialization, `ClientHandler` receives default parameter status packets from the `Supavisor.subscribe_global` function.
- `DbHandler` forwards all parameter status packets from the tenant's database to `Manager`.
- `ClientHandler` waits for up to 10 seconds to receive parameter status from `Manager`. If no packets are received within this time, it sends a "stub" packet to the client with the server version stored in the tenant's record.
- When `Manager` receives packets from `DbHandler`, it compares the values with those in the `default_parameter_status` column. If they are different, it updates them and sends a message to all of the tenant's clients to stop the connection.


lib/supavisor/helpers.ex
```diff
@@ -0,0 +1,43 @@
...
+          port: params["db_port"],
+          database: params["db_database"],
+          password: user["db_password"],
+          username: user["db_user"]
+        )
+
+      check =
+        Postgrex.query(conn, "select version()", [])
+        |> case do
+          {:ok, %{rows: [[version]]}} = r ->
```
abc3, line 19:
```suggestion
          {:ok, %{rows: [[version]]}} ->
```

lib/supavisor/tenants.ex
```diff
@@ -236,4 +238,29 @@ defmodule Supavisor.Tenants do
   def change_user(%User{} = user, attrs \\ %{}) do
     User.changeset(user, attrs)
   end
+
+  @spec get_user(String.t(), String.t() | nil) :: Ecto.Queryable.t()
```
abc3, line 242:
```suggestion
  @spec build_user_query(String.t(), String.t() | nil) :: Ecto.Queryable.t()
```

config/runtime.exs
```diff
@@ -1,8 +1,8 @@
 import Config
 
 secret_key_base =
-  if config_env() == :test do
-    "test"
+  if Enum.member?([:dev, :test], config_env()) do
```
josevalim, line 4: nitpick:

```suggestion
  if config_env() in [:dev, :test] do
```

lib/supavisor/manager.ex
```diff
@@ -99,4 +124,22 @@ defmodule Supavisor.Manager do
...
+        {:cont, nil}
+      end
+    end)
+    |> case do
+      nil ->
+        :ok
+
+      other ->
+        other
+    end
```
josevalim, line 143:
```suggestion
    Enum.find_value(ps, :ok, fn {key, value} ->
      if def_ps[key] && def_ps[key] != value do
        {:error, "Parameter #{key} changed from #{def_ps[key]} to #{value}"}
      end
    end)
```

lib/supavisor.ex
```diff
@@ -40,19 +40,20 @@ defmodule Supavisor do
...
   end
 
-  @spec subscribe_local(pid, String.t(), String.t()) :: {:ok, workers} | {:error, any()}
+  @spec subscribe_local(pid, String.t(), String.t()) :: {:ok, workers, iodata()} | {:error, any()}
   def subscribe_local(pid, tenant, user_alias) do
     with {:ok, workers} <- get_local_workers(tenant, user_alias),
-         :ok <- Manager.subscribe(workers.manager, pid) do
-      {:ok, workers}
+         {:ok, ps} <- Manager.subscribe(workers.manager, pid) do
+      Process.monitor(workers.manager)
```
josevalim, line 47: Where is the `:DOWN` for this monitor handled?

josevalim (commented):
LGTM! I have dropped only minor nits. :)

lib/supavisor/protocol/server.ex
```diff
@@ -348,4 +348,10 @@ defmodule Supavisor.Protocol.Server do
   def ready_for_query() do
     @ready_for_query
   end
+
+  @spec stub_ps(binary()) :: iodata()
+  def stub_ps(version) do
+    payload = ["server_version", <<0>>, version, <<0>>]
+    [?S, <<IO.iodata_length(payload) + 4::32>>, payload]
+  end
```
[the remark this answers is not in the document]
abc3: right, thanks for the hint!

lib/supavisor.ex
abc3: in the client connection https://github.com/supabase/supavisor/blob/68979bb6448e38bf5efe8daf502399358bd61e62/lib/supavisor/client_handler.ex#L195-L207

josevalim: So maybe it is preferrable to monitor it there? So we monitor it close to where the DOWN is received?

abc3: Yep, that would be better, thanks for finding that!

PR #107, opened 2023-06-02
Prevent linking to the db process on startup when transaction mode

abc3:
This PR fixes a bug where on startup the client_handler, in transaction mode, locks the db_handler process. This should only occur in session mode


philss (commented):
Makes sense. LGTM! :+1: 

Nitpicking: is it possible to add a test for this? I mean, if it's too complicated, I think it's OK to leave as it is.

abc3:
@philss Yes, it's a bit complicated, but I made this check through the integration test. Please take a look at the latest commit when you have a chance https://github.com/supabase/supavisor/pull/107/commits/c2f225134495e72727bcad435d847b084e2a88aa

philss:
@abc3 awesome! Normally it's not a good idea to check the genserver's internal state, but I think in these situations it's fine :)

Issue #11, opened 2023-02-02
Github Action for automatically running pgbench


chasers:

J0:
Hey

Quick check: for this issue what would happen to the logs after running `PGPASSWORD=postgres pgbench -M extended --transactions 100 --jobs 10 --client 100 -h localhost -p 7654 -U postgres.localhost postgres` or similar?

 Would it be piped somewhere or would it remain in the CI logs in case anyone wanted to inspect that changes are not affecting the performance of Supavisor too much.

If it's still relevant at all I mean

mizoz:
Hi! I'm a developer looking to contribute. I'd love to work on this — could you please assign it to me? Happy to get started right away.

anujbolewar:
Hi! ,  I'd love to take it on. Could you assign it to me?

PR #109, opened 2023-06-05
feat: add pgbench

J0:
Drive by PR to add a Github Action check -  aims to address #11

TODO:
- [ ] Figure out how to add tenant while running server in background
- [x] Rename the GH Action name


J0:
Close for now, will re-open on the weekend when I have time to look into this if no one has finished it by then

J0:
At this point, I'm stuck on getting curl to run successfully so that I can load a tenant  :/ I'll see if there's anyone I can get help from or seek help from StackOverflow.

If anyone has suggestions do let me know!

PR #115, opened 2023-06-19
chore: support ipv4 and ipv6 for tenant's db

abc3:
This PR adds the possibility of choosing the IP version for connecting to tenants' databases. It also provides an option to automatically select the IP version by checking with `:inet.gethostbyname`.


lib/supavisor.ex
```diff
@@ -2,6 +2,7 @@ defmodule Supavisor do
   @moduledoc false
   require Logger
   alias Supavisor.{Tenants, Tenants.Tenant, Manager}
+  alias Supavisor.Helpers, as: H
```
abc3, line 5:
```suggestion
  alias Supavisor.Helpers, as: H
  alias Supavisor.{Tenants, Tenants.Tenant, Manager}
```

priv/repo/migrations/20230619091028_add_tenant_ip_version.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Supavisor.Repo.Migrations.AddTenantIpVersion do
+  use Ecto.Migration
+
+  def up do
+    alter table("tenants", prefix: "_supavisor") do
+      add(:ip_version, :string, null: false)
```
philss, line 6: Have you consider adding a [constraint](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#constraint/3) to this column? I think this can better guarantee the consistency of data by checking that the value can only be `v4`, `v6` or `auto`. It's a behaviour similar to what would be if you the "enum" database type was used instead of string. WDYT?

philss: Also, it would be nice to have a default value in the DB column as well. Maybe "auto"? WDYT?

philss (commented):
LGTM! :+1: 

Just two suggestions about the db table :)

priv/repo/migrations/20230619091028_add_tenant_ip_version.exs
abc3: good call, I added these updates in the last commit https://github.com/supabase/supavisor/pull/115/commits/6fdfdbb5afa07612408ea96e86bc3fe30e6f95bc

chasers (approved):

PR #119, opened 2023-06-30
feat: add mix tasks for generating release artifacts

abc3:
This PR adds new mix tasks that facilitate the generation of release artifacts.

[Distillery's implementation ](https://github.com/bitwalker/distillery)is used for generating the `appup` file and `relup` is generated via `:systools.make_relup/4`.


lib/tasks/gen.appup.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Mix.Tasks.Supavisor.Gen.Appup do
+  @moduledoc """
+  Generates an appup file for a given release.
+
+  ## Examples
+
+    # Generate an appup from 0.0.1 to 0.0.2 versions
+
+    mix supavisor.gen.appup --from=0.0.1 --to=0.0.2
```
wojtekmach, line 9: 4 space indent for code blocks:

```suggestion
      # Generate an appup from 0.0.1 to 0.0.2 versions

      mix supavisor.gen.appup --from=0.0.1 --to=0.0.2
```

lib/tasks/gen.tarball.ex
```diff
@@ -0,0 +1,96 @@
...
+  ## Examples
+
+    # Generate a tarball for version 0.0.1
+
+    mix supavisor.gen.tarball --vsn=0.0.1
+  """
+
+  use Mix.Task
+
+  @spec run(OptionParser.argv()) :: no_return
```
wojtekmach, line 14:
```suggestion
  @impl true
```

:)

lib/tasks/gen.tarball.ex
```diff
@@ -0,0 +1,96 @@
+defmodule Mix.Tasks.Supavisor.Gen.Tarball do
```
wojtekmach, line 1: I believe the same can be achieved by adding `:tar` to release `:steps` in mix.exs?

wojtekmach (commented):
LGTM

lib/tasks/gen.tarball.ex
abc3: Oh, interesting, I'm gonna read mix manual 🤔

wojtekmach: Yup!

```elixir
  def project do
    [
      # ...
      releases: [
        foo: [
          steps: [
            :assemble,
            &foo/1,
            :tar
          ]
        ]
      ]
    ]
  end

  defp foo(release) do
    Mix.Task.run("eval", ["IO.inspect 1 + 2"])
    release
  end
```

josevalim:
Also, for completeness, we talk about the challenges for hot code upgrades here: https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-hot-code-upgrades :)

abc3:
@josevalim, thanks for pointing out this notice to me. We want to use the built-in upgrading for minor changes and when it doesn't involve state changes. Our goal is to reduce user disconnection during updates when it is possible

chasers (approved):

PR #124, opened 2023-07-05
SSL support for upstream connections

abc3:
This PR introduces support for establishing secure connections between Supavisor and a tenant's database. If you need to use client certificates, there are two methods:

- declare an environment variable `GLOBAL_UPSTREAM_CA_PATH` with the path to the certificate that will be used for all tenants
- pass the body of the certificate per tenant through the API

Here's how the extended API looks:
```js
{
  "tenant": {
    ....
    "upstream_ssl": true, // true | false
    "upstream_verify": "peer", // "none" | "peer"
    "upstream_tls_ca": "-----BEGIN CERTIFICATE-----\nblalblalblablalblalblaba\n-----END CERTIFICATE-----\n"
  }
}

```


lib/supavisor/db_handler.ex
```diff
@@ -255,4 +263,83 @@ defmodule Supavisor.DbHandler do
...
+    :ok = sock_send(sock, msg)
+    {:ok, sock}
+  end
+
+  defp sock_send({mod, sock}, data) do
+    mod.send(sock, data)
+  end
+
+  defp set_sock_opts({:gen_tcp, sock}, opts) do
+    :inet.setopts(sock, opts)
```
wojtekmach, line 339: we should check the return value, either here or at the call site:

```suggestion
    :ok = :inet.setopts(sock, opts)
```

lib/supavisor/db_handler.ex
```diff
@@ -255,4 +263,83 @@ defmodule Supavisor.DbHandler do
...
+      :ok -> ssl_recv(sock, auth)
+      {:disconnect, _, _} = dis -> dis
+    end
+  end
+
+  @spec ssl_recv(term(), map) ::
+          {:ok, {:ssl, :sslsock}}
+          | {:error, :ssl_not_available | :ssl_connect_error}
+  defp ssl_recv({:gen_tcp, sock} = s, auth) do
+    case :gen_tcp.recv(sock, 1, :infinity) do
```
wojtekmach, line 282: double-checking if infinity is safe here or instead we should, say, 15s? I think that's default timeout for stuff like this that Postgrex uses.

lib/supavisor/db_handler.ex
```diff
@@ -255,4 +263,83 @@ defmodule Supavisor.DbHandler do
...
+    end
+  end
+
+  defp ssl_connect({:gen_tcp, sock}, auth, timeout \\ 5000) do
+    opts =
+      case auth.upstream_verify do
+        :peer ->
+          [
+            verify: :verify_peer,
+            cacerts: [auth.upstream_tls_ca],
```
wojtekmach, line 303: unless you have a strong opinion about using `ca/tls_ca` names, I'd consider using `cacerts/cacertfile` names because they'd be immediately obvious to most Elixir/Erlang people.

```suggestion
            cacerts: [auth.upstream_cacerts],
```

lib/supavisor/db_handler.ex
```diff
@@ -255,4 +263,83 @@ defmodule Supavisor.DbHandler do
...
+
+  defp send_startup(sock, auth) do
+    msg =
+      :pgo_protocol.encode_startup_message([
+        {"user", auth.user},
+        {"database", auth.database},
+        {"application_name", auth.application_name}
+      ])
+
+    :ok = sock_send(sock, msg)
```
wojtekmach, line 330: double-checking if it's ok to crash or we should gracefully handle getting socket error here, e.g. `{:error, :closed}`.

lib/supavisor/helpers.ex
```diff
@@ -4,16 +4,27 @@ defmodule Supavisor.Helpers do
   @spec check_creds_get_ver(map) :: {:ok, String.t()} | {:error, String.t()}
   def check_creds_get_ver(params) do
     Enum.reduce_while(params["users"], {nil, nil}, fn user, _ ->
+      ssl_opts =
+        if params["upstream_ssl"] and params["upstream_verify"] == "peer" do
```
wojtekmach, line 8: would `params["upstream_ssl"]` be `true` or `"true"` (string)? The latter would crash `and`. I think it's better it's `true` and we use `and` but just double-checking!

lib/supavisor/helpers.ex
```diff
@@ -106,4 +117,31 @@ defmodule Supavisor.Helpers do
...
+    case :public_key.pem_decode(cert) do
+      [] ->
+        {:error, :cant_decode_certificate}
+
+      pem_entries ->
+        cert = for {:Certificate, cert, :not_encrypted} <- pem_entries, do: cert
+
+        case cert do
+          [cert] -> {:ok, cert}
+          _ -> {:error, :invalid_certificte}
```
wojtekmach, line 132:
```suggestion
          _ -> {:error, :invalid_certificate}
```

lib/supavisor/helpers.ex
```diff
@@ -106,4 +117,31 @@ defmodule Supavisor.Helpers do
...
+
+  @spec upstream_cert(binary() | nil) :: binary() | nil
+  def upstream_cert(default) do
+    case Application.get_env(:supavisor, :global_upstream_ca) do
+      nil ->
+        default
+
+      cert ->
+        cert
+    end
```
wojtekmach, line 145:
```suggestion
    Application.get_env(:supavisor, :global_upstream_ca, default)
```

lib/supavisor/monitoring/telem.ex
```diff
@@ -3,9 +3,16 @@ defmodule Supavisor.Monitoring.Telem do
 
   require Logger
 
-  @spec network_usage(atom(), port(), String.t(), String.t()) :: :ok
-  def network_usage(type, socket, tenant, user_alias) do
-    case :inet.getstat(socket) do
+  @spec network_usage(
+          atom(),
+          {:gen_tcp | :ssl, :gen_tcp.socket() | :ssl.socket()},
```
wojtekmach, line 8:
```suggestion
          {:gen_tcp, :gen_tcp.socket()} | {:ssl, :ssl.socket()},
```

lib/supavisor_web/controllers/tenant_controller.ex
```diff
@@ -82,34 +89,68 @@ defmodule SupavisorWeb.TenantController do
...
+
+            tenant ->
+              tenant = Repo.preload(tenant, :users)
+
+              with {:ok, %TenantModel{} = tenant} <-
+                     Tenants.update_tenant(tenant, params) do
+                for user <- tenant.users do
+                  Supavisor.stop(tenant.external_id, user.db_user_alias)
+                  |> then(&"Stop #{user.db_user_alias}.#{tenant.external_id}: #{inspect(&1)}")
+                  |> Logger.warning()
```
wojtekmach, line 147: nitpick: this is subjective but in this particular case I'd really consider breaking up the pipe:

```suggestion
                  result = Supavisor.stop(tenant.external_id, user.db_user_alias)
                  Logger.warning("Stop #{user.db_user_alias}.#{tenant.external_id}: #{inspect(result)}")
```

test/supavisor/db_handler_test.exs
```diff
@@ -26,8 +26,10 @@ defmodule Supavisor.DbHandlerTest do
   describe "handle_event/4" do
     test "db is avaible" do
       :meck.new(:gen_tcp, [:unstick, :passthrough])
-      :meck.expect(:gen_tcp, :connect, fn _host, _port, _socket_opts -> {:ok, :socket} end)
-      :meck.expect(:gen_tcp, :send, fn _socket, _msg -> :ok end)
+      :meck.new(:inet, [:unstick, :passthrough])
+      :meck.expect(:gen_tcp, :connect, fn _host, _port, _sock_opts -> {:ok, :sock} end)
+      :meck.expect(:gen_tcp, :send, fn _sock, _msg -> :ok end)
+      :meck.expect(:inet, :setopts, fn _sock, _opts -> :ok end)
```
wojtekmach, line 32: this is a separate discussion (apologies, I don't think we had one about it!) but while mocking is pretty convenient here I think it provides less confidence and more fragile (totally coupled to the implementation) test suite. Since it's trivial to spawn tcp sockets on the BEAM, I'd consider doing just that and testing against them.

lib/supavisor/db_handler.ex
```diff
@@ -50,26 +50,28 @@ defmodule Supavisor.DbHandler do
...
-        msg =
-          :pgo_protocol.encode_startup_message([
-            {"user", auth.user},
-            {"database", auth.database},
-            {"application_name", auth.application_name}
-          ])
+        case handshake(sock, auth) do
+          {:ok, sock} ->
+            # Back to active from here
+            set_sock_opts(sock, [{:active, true}])
```
wojtekmach, line 68: nitpick: I'd consider calling this `setopts` to match socket API, I think it will be immediately obvious. You can also use shorthand `active: true`:

```suggestion
            setopts(sock, active: true)
```

btw, since you only ever use it for socket activation you can instead do:

```
activate(sock)
```

I vaguely recall we have exactly that little helper in Postgrex.

It is totally fine as is of course!

lib/supavisor/db_handler.ex
```diff
@@ -50,26 +50,28 @@ defmodule Supavisor.DbHandler do
...
+            {:next_state, :authentication, %{data | sock: sock}}
 
-        :ok = :gen_tcp.send(socket, msg)
-        {:next_state, :authentication, %{data | socket: socket}}
+          {:error, error} ->
+            Logger.error("Handshake error #{inspect(error)}")
+            {:keep_state_and_data, {:state_timeout, 2_500, :connect}}
```
wojtekmach, line 73: consider adding a `@some_timeout 2500` at the top of the module, otherwise it's easy to miss imo.

lib/supavisor/db_handler.ex
```diff
@@ -114,11 +117,11 @@ defmodule Supavisor.DbHandler do
                 ]
 
                 bin = :pgo_protocol.encode_scram_response_message(sasl_initial_response)
-                :gen_tcp.send(data.socket, bin)
+                sock_send(data.sock, bin)
```
wojtekmach, line 120: consider checking return and gracefully handling errors. Or just crash if it wasn't successful.

```suggestion
                :ok = sock_send(data.sock, bin)
```

lib/supavisor/db_handler.ex
abc3: no, using infinity is not safe, thank you for the suggestion 👍

lib/supavisor/db_handler.ex
abc3: that looks much better 👍

lib/supavisor/db_handler.ex
abc3: yeah, good call

lib/supavisor/helpers.ex
abc3: the `upstream_ssl` is a boolean, but it's not required, so it can also be `nil`. I've added double negation
```elixir
if !!params["upstream_ssl"] and params["upstream_verify"] == "peer" do
```

lib/supavisor/db_handler.ex
abc3: I don't have a strong opinion about the naming. I just wanted to use something similar to [the PostgreSQL context](https://www.postgresql.org/docs/current/ssl-tcp.html)

Screenshot 2023-07-07 at 11 44 04

lib/supavisor/helpers.ex
abc3: 🤦

lib/supavisor/db_handler.ex
abc3: in the next PRs, I'm gonna add `downstream_tls_cert`, `downstream_tls_key`, etc. WDYT?

test/supavisor/db_handler_test.exs
abc3: that is a good point, added issue #128

lib/supavisor/db_handler.ex
wojtekmach: I see, sounds good to me!

lib/supavisor/helpers.ex
abc3: I remembered why I did like that, `supavisor.global_upstream_ca` can be declared as `nil`, and the default value via `get_env/3` will not work
https://github.com/supabase/supavisor/blob/8ff576e084083ac4449993144024be720008144a/config/runtime.exs#L100-L119

wojtekmach: I see, that makes sense. In that case you can also do:

```elixir
Application.get_env(:supavisor, :global_upstream_ca) || default
```

abc3:
@wojtekmach Can you please take a look at the new changes when you have a chance?

config/runtime.exs
```diff
@@ -95,6 +97,16 @@ config :libcluster,
   debug: false,
   topologies: topologies
 
+upstream_ca =
+  if System.get_env("GLOBAL_UPSTREAM_CA_PATH") do
+    File.read!(System.get_env("GLOBAL_UPSTREAM_CA_PATH"))
```
wojtekmach, line 102:
```suggestion
  if path = System.get_env("GLOBAL_UPSTREAM_CA_PATH") do
    File.read!(path)
```

wojtekmach (commented):
LGTM!

kiwicopple (approved):

PR #126, opened 2023-07-06
Add SSL to DbHandler

acco:


lib/supavisor/db_handler.ex
```diff
@@ -53,23 +53,25 @@ defmodule Supavisor.DbHandler do
     socket_opts = [
       :binary,
       {:packet, :raw},
-      {:active, true},
+      {:active, false},
```
acco, line 56: Start off in not-active mode. Allows us to pull data out of socket for SSL (vs writing a bunch of receivers for responses). Mirrors what Postgrex does.

lib/supavisor/db_handler.ex
```diff
@@ -53,23 +53,25 @@ defmodule Supavisor.DbHandler do
...
       auth.ip_version
     ]
 
     case :gen_tcp.connect(auth.host, auth.port, socket_opts) do
       {:ok, socket} ->
-        Logger.debug("auth #{inspect(auth, pretty: true)}")
+        sock = {:gen_tcp, socket}
```
acco, line 62: Same format Postgrex uses ({module name, socket})

lib/supavisor/db_handler.ex
```diff
@@ -82,7 +84,8 @@ defmodule Supavisor.DbHandler do
     {:keep_state_and_data, {:next_event, :internal, :connect}}
   end
 
-  def handle_event(:info, {:tcp, _, bin}, :authentication, data) do
+  def handle_event(:info, {tcp_or_ssl, _, bin}, :authentication, data)
+      when tcp_or_ssl in [:tcp, :ssl] do
```
acco, line 88: Another way to do this: have a `handle_event` for each of `:tcp` and `:ssl` and route to the same place.

lib/supavisor/db_handler.ex
```diff
@@ -255,4 +264,72 @@ defmodule Supavisor.DbHandler do
 
     :keep_state_and_data
   end
+
+  # Adapted from Postgrex.Protocol
+  # TODO: Pull over handshake timeout from Postgrex.Protocol?
+  defp handshake(sock, %{ssl: true} = auth), do: ssl(sock, auth)
+  defp handshake(sock, %{ssl: false} = auth), do: send_startup(sock, auth)
```
acco, line 271: This mirrors a lot from Postgrex, except doesn't have some more advanced stuff that it does, like a handshake timeout

PR #127, opened 2023-07-06
Validate startup packets

acco:
We get a lot of log messages about `function_clause` errors etc that are actually just a result of bots sending packets to Supavisor.

This change filters out the noise from unserious clients (like bots/scrapers) sending requests. It does so by validating the startup packet is in the rough shape we're expecting.


hauleth:
This one got resolved in different commit https://github.com/supabase/supavisor/commit/4265b5143a4577f064b5fb1d7b23c6cd6bc36c18#diff-3ca812e1c5954a2b89bbcc2c56e525a015e4eeb867abbc270f613f13ee94215cR105.

PR #130, opened 2023-07-11
Option to allow only SSL connection

abc3:
This PR introduces the ability to support only SSL connections.

An additional boolean option `enforce_ssl` has been added to the tenant's API. By default, this is set to `false`.
If `enforce_ssl` is set to `true` and the client doesn't initiate an SSL connection, Supavisor will terminate it, returning the error: `FATAL: SSL connection is required`.


philss (commented):
LGTM! :+1:

PR #133, opened 2023-07-18
feat: retrieve the tenant id from the server name indication

abc3:
This PR introduces a new method to retrieve the tenant ID. If the client uses an SSL connection, we attempt to find the tenant by using `sni_hostname`

### DIFF
--- a/priv/repo/migrations/20230718175315_add_sni_host.exs
+++ b/priv/repo/migrations/20230718175315_add_sni_host.exs
@@ -1,15 +1,9 @@
 defmodule Supavisor.Repo.Migrations.AddSniHost do
   use Ecto.Migration
 
-  def up do
+  def change do
     alter table("tenants", prefix: "_supavisor") do
       add(:sni_hostname, :string, null: true)
-    end
-  end
-
-  def down do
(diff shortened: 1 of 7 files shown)


priv/repo/migrations/20230718175315_add_sni_host.exs
```diff
@@ -0,0 +1,15 @@
...
+    alter table("tenants", prefix: "_supavisor") do
+      add(:sni_hostname, :string, null: true)
+    end
+  end
+
+  def down do
+    alter table("tenants", prefix: "_supavisor") do
+      remove(:sni_hostname)
+    end
+  end
```
philss, line 14: I think you can write this migration with only a `change` function:

```suggestion
  def change do
    alter table("tenants", prefix: "_supavisor") do
      add(:sni_hostname, :string, null: true)
    end
  end
```

WDYT?

philss (commented):
Only a tiny suggestion, but LGTM! :+1:

priv/repo/migrations/20230718175315_add_sni_host.exs
abc3: looks good, thanks!

PR #137, opened 2023-07-21
fix: update the query for retrieving user creds

abc3:
This PR has updated the function for retrieving user credentials. If the username contains `tenant_id`, the query will attempt to use it. If not, it will use `sni_hostname` instead.


lib/supavisor/tenants.ex
```diff
@@ -45,6 +45,10 @@ defmodule Supavisor.Tenants do
 
   @spec get_user(String.t(), String.t() | nil, String.t() | nil) ::
           {:ok, map()} | {:error, :not_found | :multiple_results}
+  def get_user(_, nil, nil) do
+    {:error, "Either external_id or sni_hostname must be provided"}
```
josevalim, line 49: This error does not match the spec. :)

lib/supavisor/tenants.ex
```diff
@@ -269,32 +243,25 @@ defmodule Supavisor.Tenants do
...
+    )
+    |> where(^with_tenant(external_id, sni_hostname))
+  end
+
+  defp with_tenant(nil, sni_hostname) do
+    dynamic([_, t], t.sni_hostname == ^sni_hostname)
+  end
+
+  defp with_tenant(external_id, _) do
+    dynamic([_, t], t.external_id == ^external_id)
```
josevalim, line 265: I believe this also works but they all work the same. :)

```suggestion
    [sni_hostname: sni_hostname]
  end

  defp with_tenant(external_id, _) do
    [external_id: external_id]
```

josevalim (commented):
LGTM, just one tiny comment on the typespec. :)

lib/supavisor/tenants.ex
abc3: 🤦

delgado3d (approved):

PR #141, opened 2023-07-26
fix: use username from the connection in process names

abc3:
The process id should be a tuple consisting of the tenant and username received from the incoming connection. This PR fixes the behavior where the process id was made up of the username from the auth_query


lib/supavisor/client_handler.ex
```diff
@@ -164,17 +165,25 @@ defmodule Supavisor.ClientHandler do
...
+         {:next_event, :internal, :subscribe}}
     end
   end
 
   def handle_event(:internal, :subscribe, _, %{tenant: tenant, user_alias: db_alias} = data) do
     Logger.info("Subscribe to tenant #{inspect({tenant, db_alias})}")
 
+    conn_user =
+      if data.proxy_type == :auth_query do
+        elem(data.auth_secrets, 1).().user
```
wojtekmach, line 178: nitpick: for the sake of being consistent I'd change this to be the same as in https://github.com/supabase/supavisor/pull/141/files#diff-2f00718c6abf75c95277a9659fea6e72650be27b36013905af1837e77af76a91R104

```suggestion
        {_method, secrets} = data.auth_secrets
        secrets.().user
```

if we don't care about method that much another option is to keep it separately as `data.auth_method` and `data.auth_secrets`.

I think it's totally fine to leave this as is, of course.

wojtekmach (commented):
LGTM!

lib/supavisor/client_handler.ex
abc3: that looks better, thanks!

delgado3d (approved):

PR #142, opened 2023-07-28
feat: add region and instance id to logger

abc3:
This PR adds the `region` and `instance id` to the logger metadata for the client/db handlers and manager process.


abc3:
Hey @supabase/dashbit, is there an easy way to globally set up a value in metadata for all processes? For example, it would be nice to see the region and instance id values in all log messages

josevalim:
Starting from Elixir v1.15, you should be able to set metadata in your config/runtime.exs:

    config :logger, :metadata, [...]

abc3:
Wow, nice! But we cannot upgrade to v1.15 because one of our dependencies doesn't support this version. We'll try to upgrade it first

josevalim:
Which dep is problematic?

Anyway, you can also do this in your application.ex:

    primary_config = :logger.get_primary_config()
    :ok = :logger.set_primary_config(:metadata, Enum.into([...yourmetadata...], primary_config.metadata))

It requires Erlang/OTP 24.

abc3:
> Which dep is problematic?

the Logflare Erlang client uses Etso, which doesn't work with Erlang 26 due to an [old Ecto version](https://github.com/evadne/etso/blob/cae311eedf42003ddbc6796798a67193d5e56454/mix.exs#L30)
```elixir
== Compilation error in file lib/ecto/query.ex ==
** (Kernel.TypespecError) lib/ecto/query.ex:430: type dynamic/0 is a built-in type and it cannot be redefined
    (elixir 1.15.4) lib/kernel/typespec.ex:961: Kernel.Typespec.compile_error/2
    (stdlib 5.0.2) lists.erl:1599: :lists.foldl_1/3
    (elixir 1.15.4) lib/kernel/typespec.ex:226: Kernel.Typespec.translate_typespecs_for_module/2
```

> Anyway, you can also do this in your application.ex:
> 
> primary_config = :logger.get_primary_config()
> :ok = :logger.set_primary_config(:metadata, Enum.into([...yourmetadata...], primary_config.metadata))
> 
> It requires Erlang/OTP 24.

That works perfectly, thanks! 🔥

lib/supavisor/client_handler.ex
```diff
@@ -107,6 +107,7 @@ defmodule Supavisor.ClientHandler do
     hello = decode_startup_packet(bin)
     Logger.debug("Client startup message: #{inspect(hello)}")
     {user, external_id} = parse_user_info(hello.payload["user"])
+
```
abc3, line 110:
```suggestion
```

lib/supavisor/db_handler.ex
```diff
@@ -30,6 +30,7 @@ defmodule Supavisor.DbHandler do
   @impl true
   def init(args) do
     Process.flag(:trap_exit, true)
+
```
abc3, line 33:
```suggestion
```

lib/supavisor/manager.ex
```diff
@@ -46,6 +46,7 @@ defmodule Supavisor.Manager do
     }
 
     Logger.metadata(project: args.tenant, user: args.user_alias)
+
```
abc3, line 49:
```suggestion
```

chasers (approved):

Issue #134, opened 2023-07-19
Limit client connections


chasers:
- [x] `default_max_clients` on tenant to use when we don’t have a user record - default `1000`
- [x] `max_clients` on user and use this when we have the user 
- [x] if null don't limit
- [x] Error message should be "Max client connections reached"

PR #144, opened 2023-08-01
feat: limit client connections

abc3:
This PR introduces a feature to limit client connections per tenant

- added `default_max_clients` as a REST API parameter for the tenant object. This sets a default limit on client connections with a default value of `1000`

- added `max_clients` as a REST API parameter for the user object. If this parameter is set to `null`, no connection limit will be applied for the user (works only through direct connection)

- if a tenant reaches their connection limit, the system will terminate the connection and display the message "Max client connections reached"

issue: https://github.com/supabase/supavisor/issues/134


lib/supavisor/manager.ex
```diff
@@ -54,15 +55,24 @@ defmodule Supavisor.Manager do
...
   def handle_call({:subscribe, pid}, _, %{tenant: tenant, user_alias: user_alias} = state) do
     Logger.info("Subscribing #{inspect(pid)} to tenant #{inspect({tenant, user_alias})}")
-    :ets.insert(state.tid, {Process.monitor(pid), pid, now()})
 
-    case state.parameter_status do
-      [] ->
-        {:reply, {:ok, []}, update_in(state.wait_ps, &[pid | &1])}
+    # don't limit if max_clients is null
+    {reply, new_state} =
+      if :ets.info(state.tid, :size) < state.max_clients do
```
josevalim, line 61: Keep in mind that, if you ever set `decentralized_counters` to true, then computing the size is slower as it has to aggregate the size from multiple cores. Right now, you don't set this option and given this is a table per manager, I don't think there would ever be a need to set, so it is not a concern. In any case, you may want to consider setting either `read_concurrency` or `write_concurrency` in the ETS table. :)

josevalim: I think a `cond` would work nicely here too to remove the nesting:

```
cond do
  :ets.info(state.tid, :size) >= state.max_clients ->
    {{:error, :max_clients_reached}, state}

  state.parameter_status == [] ->
    {{:ok, []}, update_in(state.wait_ps, &[pid | &1])}

  true ->
    {{:ok, state.parameter_status}, state}
end
```

josevalim (commented):
LGTM!

lib/supavisor/manager.ex
abc3: Without an additional level of nesting, we would have to call `:ets.insert/2` multiple times. The updated part of the code would look like this:
```elixir
cond do
  :ets.info(state.tid, :size) >= state.max_clients ->
    {{:error, :max_clients_reached}, state}

  state.parameter_status == [] ->
    :ets.insert(state.tid, {Process.monitor(pid), pid, now()})
    {{:ok, []}, update_in(state.wait_ps, &[pid | &1])}

  true ->
    :ets.insert(state.tid, {Process.monitor(pid), pid, now()})
    {{:ok, state.parameter_status}, state}
end
```

existing:
```elixir
if :ets.info(state.tid, :size) < state.max_clients do
  :ets.insert(state.tid, {Process.monitor(pid), pid, now()})

  case state.parameter_status do
    [] ->
      {{:ok, []}, update_in(state.wait_ps, &[pid | &1])}

    ps ->
      {{:ok, ps}, state}
  end
else
  {{:error, :max_clients_reached}, state}
end
```

Let me know if the example with "cond" looks better than the existing one 🙏

delgado3d (approved):

jakeleventhal:
@abc3 If the user reaches the connection limit, why not allow them to queue the connection (governed by connection timeout in the connection string)

PR #145, opened 2023-08-03
feat: api endpoint to terminate tenancy tree connection

abc3:
This PR adds a `/tenants/:tenant_id/terminate` endpoint, which searches the entire cluster for the tenant's connections by going through all dynamic supervisor shards and stopping these PIDs.

### DIFF
--- a/lib/supavisor.ex
+++ b/lib/supavisor.ex
@@ -86,7 +86,7 @@ defmodule Supavisor do
   """
   @spec dirty_terminate(String.t(), pos_integer()) :: map()
   def dirty_terminate(tenant, timeout \\ 15_000) do
-    Registry.lookup(Supavisor.Registry.TenantsSups, tenant)
+    Registry.lookup(Supavisor.Registry.TenantSups, tenant)
     |> Enum.reduce(%{}, fn {pid, %{user: user}}, acc ->
       stop =
         try do
--- a/lib/supavisor/application.ex
+++ b/lib/supavisor/application.ex
@@ -60,7 +60,7 @@ defmodule Supavisor.Application do
       PromEx,
       {Registry, keys: :unique, name: Supavisor.Registry.Tenants},
       {Registry, keys: :unique, name: Supavisor.Registry.ManagerTables},
-      {Registry, keys: :duplicate, name: Supavisor.Registry.TenantsSups},
+      {Registry, keys: :duplicate, name: Supavisor.Registry.TenantSups},
       {Cluster.Supervisor, [topologies, [name: Supavisor.ClusterSupervisor]]},
       Supavisor.Repo,
       # Start the Telemetry supervisor
--- a/lib/supavisor/tenant_supervisor.ex
+++ b/lib/supavisor/tenant_supervisor.ex
@@ -33,7 +33,7 @@ defmodule Supavisor.TenantSupervisor do
       {Manager, args}
     ]
 
-    Registry.register(Supavisor.Registry.TenantsSups, args.tenant, %{user: args.user_alias})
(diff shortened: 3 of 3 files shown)


lib/supavisor_web/router.ex
```diff
@@ -46,10 +46,11 @@ defmodule SupavisorWeb.Router do
   scope "/api", SupavisorWeb do
     pipe_through(:api)
 
-    get("/tenants", TenantController, :index)
+    # get("/tenants", TenantController, :index)
```
abc3, line 49:
```suggestion
```

test/supavisor_web/controllers/tenant_controller_test.exs
```diff
@@ -103,13 +103,13 @@ defmodule SupavisorWeb.TenantControllerTest do
...
-      assert is_list(data)
-    end
-  end
+  # describe "list tenants" do
+  #   test "renders tenants", %{conn: conn} do
+  #     conn = get(conn, Routes.tenant_path(conn, :index))
+  #     data = json_response(conn, 200)["data"]
+  #     assert is_list(data)
+  #   end
+  # end
```
abc3, line 113:
```suggestion
```

lib/supavisor_web/controllers/tenant_controller.ex
```diff
@@ -173,4 +173,23 @@ defmodule SupavisorWeb.TenantController do
 
     send_resp(conn, code, "")
   end
+
+  def terminate(conn, %{"external_id" => external_id}) do
+    Logger.metadata(project: external_id)
+
+    result =
+      [node(), :"node2@127.0.0.1" | Node.list()]
```
abc3, line 181:
```suggestion
      [node() | Node.list()]
```

chasers (approved):

lib/supavisor.ex
```diff
@@ -80,6 +80,54 @@ defmodule Supavisor do
...
 
+  @doc """
+  During netsplits, or due to certain internal conflicts, :syn may store inconsistent data across the cluster.
+  This function terminates all connection trees related to a specific tenant.
+  It accomplishes this by going through all dynamic supervisor shards and searching for all supervisors
+  """
+  @spec dirty_terminate(String.t(), pos_integer()) :: map()
+  def dirty_terminate(tenant, timeout \\ 15_000) do
+    :erlang.whereis(Supavisor.DynamicSupervisor)
+    |> :supervisor.which_children()
```
josevalim, line 91: I am worried this request and traversal can be expensive when running thousands of clients. Maybe you should setup another registry (this one with duplicate keys) that stores all connections for a given tenant?

josevalim: Btw, you want to move these registries to the supervision tree:

https://github.com/supabase/supavisor/blob/d90556bbf8bc8971e9f940370edf52650badcdbb/lib/supavisor/application.ex#L55C1-L64C1

:)

It should be `{Registry, options}`. :)

josevalim: And to be clear, you won't use the `{:via, ...}` API to register those, but rather call Registry.register(...) directly to store all children for a tenant.

abc3: My intention was to get supervisors from OTP internals to ensure that nothing was lost. But yes, it can be a heavy operation when there are a lot of connection trees. 

Take a look, please, at the updated commit when you have free time to ensure that I understand you correctly https://github.com/supabase/supavisor/pull/145/commits/1afe81a9720f9a1d8b4fde6b1345cd39aef92fa1

josevalim: That commit is perfect. 💯  :)

abc3: Thanks for such a quick review!

Issue #123, opened 2023-07-04
Metric endpoint per tenant


chasers:
We need to expose a metric endpoint per tenant where all the metrics are only for a certain tenant.
- [x] `/metrics/:external_id`
- [x] filter per tenant metrics out of the `/metrics` endpoint so it stays small
- [x] prioritize these metrics:
<img width="746" alt="Screenshot 2023-08-07 at 9 37 43 AM" src="https://github.com/supabase/supavisor/assets/1019814/15e51c1a-e765-4f1a-96f4-a708b5491048">

PR #147, opened 2023-08-07
feat: metric endpoint per tenant

abc3:
details: https://github.com/supabase/supavisor/issues/123


lib/supavisor/monitoring/prom_ex.ex
```diff
@@ -82,6 +82,31 @@ defmodule Supavisor.Monitoring.PromEx do
...
+  def do_cache_tenants_metrics() do
+    metrics = get_metrics()
+
+    Registry.select(Supavisor.Registry.TenantSups, [{{:"$1", :_, :_}, [], [:"$1"]}])
+    |> Enum.uniq()
+    |> Enum.each(fn tenant ->
+      filtered =
+        metrics
+        |> String.split("\n")
+        |> Enum.filter(&String.contains?(&1, "tenant=\"#{tenant}\""))
```
chasers, line 94: I'd probably Stream where you can as this list could get very large.

lib/supavisor/monitoring/prom_ex.ex
```diff
@@ -82,6 +82,30 @@ defmodule Supavisor.Monitoring.PromEx do
...
+  @spec do_cache_tenants_metrics() :: :ok
+  def do_cache_tenants_metrics() do
+    metrics = get_metrics() |> String.split("\n")
+
+    Registry.select(Supavisor.Registry.TenantSups, [{{:"$1", :_, :_}, [], [:"$1"]}])
+    |> Enum.uniq()
+    |> Enum.each(fn tenant ->
+      filtered =
+        metrics
+        |> Stream.filter(&String.contains?(&1, "tenant=\"#{tenant}\""))
```
philss, line 94: I believe `Stream` is not going to make much of a difference here, because you are immediately running the stream with one operation. It's fine to keep the `Stream` module, but `Enum` should be more clear here. WDYT?

I see a possible performance issue, because you are traversing the `metrics` list multiple times (the entire list for each tenant). If the list of metrics is not big enough, then this is fine. But otherwise, it would be nice to reduce the list for each tenant interaction (maybe Enum.reduce + Enum.split_with).

Another alternative would be to categorize each line by capturing the tenant name - using regex and named captures -, and then group the lines by tenant. I'm not entirely sure if this way is going to be faster, but could be an alternative here.

WDYT? If the lists are not big enough, you can ignore me :)

philss (commented):
LGTM! :+1: 

I added a minor suggestion, but it would depend if the size of metrics is big enough.

lib/supavisor/monitoring/prom_ex.ex
abc3: Good call! Can you please take another look at the updated function?https://github.com/supabase/supavisor/pull/147/commits/452c6e41c5fb7f8de89184c3234c21d7c4a47d2d

chasers: @philss yeah this concern is because right now our Realtime metrics endpoint is 20MB and we need to add metrics there. Metrics per tenant explode the total count so that list of metrics can get very large very easily.

chasers: Point is ... whatever you think is fine that was just my concern.

chasers (approved):

lib/supavisor/monitoring/prom_ex.ex
philss: @chasers oh, I see. If think this solution is good for now, but may not be good for the long run.

I'm going to investigate if we could query directly PromEx and avoid this filtering in memory, against strings.
For what I could understand, we cannot. But I'm going to double check.

abc3: another possible solution is to run PromEx per tenant, but it will take some time to implement, and it's unclear how to grab these metrics for all tenants

philss: > another possible solution is to run PromEx per tenant, but it will take some time to implement, and it's unclear how to grab these metrics for all tenants at once

Yeah, this could be a good way to solve the problem! I'm not sure if PromEx is adapted for this kind of usage, so it could require modifications to the lib.

> I'm going to investigate if we could query directly PromEx and avoid this filtering in memory, against strings.
For what I could understand, we cannot. But I'm going to double check.

After double checking, I can confirm that we cannot do that unless we use private APIs.

I believe a small performance / memory usage improvement could be to make the [`get_metrics()`](https://github.com/supabase/supavisor/blob/2540d0ad5e80149bde428faa6158b787f6d09531/lib/supavisor/monitoring/prom_ex.ex#L67) function to return a list of metrics instead of the the joined string. This way you would only join once per tenant, and wouldn't need to split like you do on line 85. The idea is to have fewer string creation, because this is expensive in OTP. WDYT?

chasers: Yeah okay I think we're going to kick this can down the road a little and optimize later. 

Would like to spend some time researching if we can use PromEx per tenant because that seems like it would be nice.

PR #155, opened 2023-08-14
fix: the total count of connected clients for a tenant

abc3:


lib/supavisor/client_handler.ex
```diff
@@ -114,6 +114,8 @@ defmodule Supavisor.ClientHandler do
 
     case Tenants.get_user(user, external_id, sni_hostname) do
       {:ok, info} ->
+        Registry.register(Supavisor.Registry.TenantClients, {info.tenant.external_id, user}, nil)
```
abc3, line 117:
```suggestion
        Registry.register(Supavisor.Registry.TenantClients, {info.tenant.external_id, user},[])
```

chasers (approved):

PR #160, opened 2023-08-24
feat: add a mode type to the pool id

abc3:
This PR introduces the capability to launch multiple pools with different types for the same user. Now, the pool id will be represented as a tuple consisting of the tenant, username and mode type (session or transaction). 

This PR also includes bug fixes and refactoring


lib/supavisor/client_handler.ex
```diff
@@ -6,18 +6,15 @@ defmodule Supavisor.ClientHandler do
...
   require Logger
 
   @behaviour :ranch_protocol
   @behaviour :gen_statem
 
   alias Supavisor.DbHandler, as: Db
-  alias Supavisor.Helpers, as: H
-  alias Supavisor.{Tenants, Protocol.Server, Monitoring.Telem}
+  alias Supavisor, as: S
+  alias S.Helpers, as: H
```
abc3, line 16:
```suggestion
  alias Supavisor.Helpers, as: H
  alias Supavisor, as: S

```
```suggestion
  alias Supavisor, as: S
  alias S.Helpers, as: H
```
```suggestion
  alias Supavisor.Helpers, as: H
  alias Supavisor, as: S
```

lib/supavisor/manager.ex
```diff
@@ -3,11 +3,10 @@ defmodule Supavisor.Manager do
   use GenServer, restart: :transient
   require Logger
 
-  alias Supavisor.PromEx.Plugins.Tenant, as: M
   alias Supavisor.Protocol.Server
   alias Supavisor.Tenants
 
-  @check_timeout 120_000
+  @check_timeout 10_000
```
abc3, line 9:
```suggestion
  @check_timeout 120_000
```

lib/supavisor.ex
```diff
@@ -104,37 +111,63 @@ defmodule Supavisor do
...
+    [node() | Node.list()]
+    |> Task.async_stream(
+      fn node ->
+        %{node => :rpc.call(node, Supavisor, :dirty_terminate, [tenant], 60_000)}
+      end,
+      timeout: :infinity
+    )
+    |> Enum.reduce([], fn resp, acc ->
+      [inspect(resp) | acc]
+    end)
```
josevalim, line 124: This is completely fine, but you may also consider using https://www.erlang.org/doc/man/erpc#multicall-4 - which will separate the separate calls without spawning new tasks. Also note that `Task.async_stream` has a concurrency limit, which I don't think you care here?

Makefile
```diff
@@ -24,8 +24,10 @@ dev.node2:
 	REGION=eu \
 	SECRET_KEY_BASE="dev" \
 	CLUSTER_POSTGRES="true" \
+	PROXY_PORT_SESSION="5442" \
+	PROXY_PORT_TRANSACTION="6553" \
```
josevalim, line 28: Notice this env var is already set earlier.

config/test.exs
```diff
@@ -6,6 +6,7 @@ config :supavisor,
   api_jwt_secret: "dev",
   metrics_jwt_secret: "dev",
   jwt_claim_validators: %{},
+  proxy_port_session: System.get_env("PROXY_PORT_TRANSACTION", "7653") |> String.to_integer(),
```
josevalim, line 9: Apologies, I may be missing something obvious, but is this config used anywhere? Also, the env var does not match the config key, I am not sure if it is intentional.

josevalim (commented):
I have added some minor nits, but it looks great to me!

My understanding is that we already had two ports before, but they used the same pool. Now they are different pool. Some other important characteristics:

1. Pools are started lazily. This means if you run both session and transaction, but only ever use one, then only one is ever created (please confirm if my understanding is correct)

2. However, connections are started eagerly, so if you are using both, then you get the double of connections, but that's most likely what you want anyway

I can see people wanting different sizes for each pool, but that can be a conversation for another day. :)

lib/supavisor.ex
abc3: good call 👍

Makefile
abc3: oops

config/test.exs
abc3: 🤦

abc3:
@josevalim sorry for the late response

> Pools are started lazily. This means if you run both session and transaction, but only ever use one, then only one is ever created (please confirm if my understanding is correct)

Yes, correct

> However, connections are started eagerly, so if you are using both, then you get the double of connections, but that's most likely what you want anyway

Yes, exactly!

> I can see people wanting different sizes for each pool, but that can be a conversation for another day. :)

Hmm, that is a good suggestion! I will bring that idea to the team 👍

lib/supavisor.ex
abc3: updated https://github.com/supabase/supavisor/pull/160/commits/a6614b363ba4c481d329cd54d32a70cca0596072

chasers (approved):

PR #162, opened 2023-08-31
[WIP] feat: add support for read replicas

abc3:


abc3:
@supabase/dashbit Could you please take a look at this PR? 🙏 
I'm still debugging and updating tests, but it's ready for review

lib/supavisor.ex
```diff
@@ -124,11 +125,22 @@ defmodule Supavisor do
...
+    body = [{{:"$2", :"$3"}}]
+
+    case Registry.select(@registry, [{match, [], body}]) do
+      [{pool, _}] ->
+        pool
+
+      [_ | _] = pools ->
+        Enum.reduce(pools, %{write: [], read: []}, fn {pid, type}, acc ->
+          update_in(acc[type], &[pid | &1])
+        end)
```
josevalim, line 140: A `Enum.group_by` may work here too.

lib/supavisor.ex
```diff
@@ -124,11 +125,22 @@ defmodule Supavisor do
     }
   end
 
-  @spec get_local_pool(id) :: pid | nil
+  @spec get_local_pool(id) :: map | pid | nil
```
josevalim, line 128: Would it be more consistent to always return a map (which may also be empty?).

lib/supavisor/client_handler.ex
```diff
@@ -478,23 +544,34 @@ defmodule Supavisor.ClientHandler do
     end
   end
 
-  @spec db_checkout(:on_connect | :on_query, map()) :: pid() | nil
-  defp db_checkout(_, %{mode: :session, db_pid: db_pid}) when is_pid(db_pid) do
+  @spec db_checkout(:write | :read | nil, :on_connect | :on_query, map) :: {pid, pid} | nil
```
josevalim, line 547: Instead of `nil`, consider using `:both`, it may be a bit easier to find in case of errors.

josevalim:
I have dropped an initial comment but I want to do a more in-depth review tomorrow. Quick question: what happens with transactions? I guess transactions always go to the write replicate?

> If Supavisor receives a read_only_sql_transaction error from the read replica backend, the query is resent to the write replica.

When could this happen?

lib/supavisor.ex
josevalim: I guess though you want to distinguish if it is a read-write cluster or a "shared" one. So perhaps the current return type is fine.

lib/supavisor/db_handler.ex
```diff
@@ -232,19 +233,53 @@ defmodule Supavisor.DbHandler do
...
-    :ok = Client.client_call(data.caller, bin, ready)
+  def handle_event(:info, {_proto, _, bin}, _, %{replica_type: :read} = data) do
+    pkts = Server.decode(bin)
+
+    resp =
+      cond do
+        Server.has_read_only_error?(pkts) ->
+          Logger.error("read only error")
+
+          if length(pkts) == 1 do
```
josevalim, line 244:
```suggestion
          with [_] <- pkts do
```

lib/supavisor/db_handler.ex
```diff
@@ -379,4 +414,14 @@ defmodule Supavisor.DbHandler do
       auth.secrets.().user
     end
   end
+
+  @spec receive_ready_for_query() :: :ok | :timeout_error
+  defp receive_ready_for_query() do
+    receive do
+      {_proto, _socket, <<?Z, 5::32, ?I>>} ->
```
josevalim, line 421: I wonder if it is guaranteed that it will always be an individual message and if part of the TCP stack may concatenate this with other messages. I would definitely add logging to the `after` branch (although it will crash anyway, so maybe it is fine).

lib/supavisor/protocol/client.ex
```diff
@@ -0,0 +1,157 @@
...
+
+      tag ->
+        payload_len = pkt_len - 4
+        <<bin_payload::binary-size(payload_len), rest2::binary>> = rest
+
+        {:ok,
+         %Pkt{
+           tag: tag,
+           len: pkt_len + 1,
+           payload: decode_payload(tag, bin_payload),
```
josevalim, line 53: Right now it seems you only use the payload, perhaps it is worth simplifying or do you plan to use further functionality?

lib/supavisor/protocol/client.ex
```diff
@@ -0,0 +1,157 @@
...
+
+      _ ->
+        nil
+    end
+  end
+
+  def decode_payload(:simple_query, payload) do
+    case String.split(payload, <<0>>) do
+      [query, ""] -> query
+      _ -> :undefined
```
josevalim, line 96: What happens when you pass `:undefined` or `nil` to Rust, which is expecting a string?

lib/supavisor.ex
```diff
@@ -156,79 +168,100 @@ defmodule Supavisor do
...
+    case type do
+      :single -> T.get_pool_config(tenant, user)
+      :cluster -> T.get_cluster_config(tenant, user)
+    end
+    |> case do
+      [_ | _] = replicas ->
+        opts =
+          Enum.map(replicas, fn replica ->
+            # cluster config has "type" field, pool config doesn't
+            if Map.has_key?(replica, :type) do
```
josevalim, line 185: I would suggest to pattern match on the type instead. It is probably more reliable than checking on a field;

```
case replica do
  %ClusterTenant{tenant: tenant, type: type} -> Map.put(tenant, :replica_type, type)
  %Tenant{...} = tenant -> Map.put(tenant, :replica_type, :write)
```

Alternatively make it so both get_cluster_config and get_pool_config return the same data structure, already pre-normalized.

lib/supavisor.ex
abc3: Yep, that was my intention. The problem is I'm trying to implement such a massive update without rewriting all the existing code which can make some areas appear odd and unclear

lib/supavisor/db_handler.ex
abc3: after this error message, the pg backend only sends a `ready_for_query` message

lib/supavisor/protocol/client.ex
abc3: I wrote this function in the beginning when I was learning pg protocol. I'll change it to something that returns only the payload 👍

lib/supavisor/protocol/client.ex
abc3: It will raise the exception:
```elixir
iex(node1@127.0.0.1)3> Supavisor.PgParser.statement_types nil
** (ArgumentError) argument error
    (supavisor 0.9.5) Supavisor.PgParser.statement_types(nil)
```
Perhaps in that case, I should send the query to the write replica and not crash the process

abc3:
> what happens with transactions? I guess transactions always go to the write replicate?

Yes, transactions are always routed to write a replica

>> If Supavisor receives a read_only_sql_transaction error from the read replica backend, the query is resent to the write replica.
> When could this happen?

For example, if a user defined function inserts or updates anything under the hood

lib/supavisor.ex
josevalim: If it helps, something that I like to do in such cases is to do code breaking changes first and rewrite the usage without adding new features. Only then I would add the new features.

Of course, it all depends on you finding out that the API is going to change early in the process, which is not always the case. In this case, it is probably not worth doing it, since the work is well underway. :)

abc3: That's a very smart approach 🤔🤔🤔

chasers (approved):

PR #165, opened 2023-09-14
feat: add client_idle_timeout option

abc3:
This PR adds a configurable option to prevent zombie issues and enables keepalive in incoming connections.

---------

- [x] If a client stays connected but doesn't send any messages over the connection for N milliseconds, then disconnect the client
- [x] Disabled by default
- [x] Put on tenant record 
- [x] Name it `client_idle_timeout` ~~or `idle_timeout_downstream`??~~
- [x] Set keepalive in ranch to true


chasers:
Let’s call this the same thing as pgbouncer `client_idle_timeout`

abc3:
the terms downstream/upstream offer more specific and clear definitions and we are already using them in the project

chasers:
Yeah but people will be looking for equivalent configs.

josevalim:
:+1: on `client_idle_timeout`.

The current implementation is great as is, because it seems you only want to enable the timeout in certain scenarios. Although I believe the current implementation would not close the connection if it asks for a query and then forgets (or is too slow) to read it for some reason, right?

In any case, if you wanted to enable the timeout on most `handle_event` returns, then I am not the biggest fan of `gen_statem`/`gen_server` timeouts because if you forget to return `timeout` in a single callback, then it may no longer trigger as expected.

abc3:
> Although I believe the current implementation would not close the connection if it asks for a query and then forgets (or is > too slow) to read it for some reason, right?

Yeah, that is not covered

>In any case, if you wanted to enable the timeout on most handle_event returns, then I am not the biggest fan of gen_statem/gen_server timeouts because if you forget to return timeout in a single callback, then it may no longer trigger as expected.

@josevalim, what would you recommend using for timeouts? Should I use `Process.send_after`? I'll need to add a bunch of other timeouts soon

josevalim:
It really depends on the nature of timeouts. Sometimes send_after. Sometimes you can store timestamps on the state and compare when messages arrive, etc. Once you have more details, let us know, and we will gladly discuss :)

delgado3d (approved):

PR #173, opened 2023-10-04
feat: add an option to set the pool's strategy, default is fifo

abc3:
This PR introduces the option to choose different queue strategies for each tenant's db pool, with FIFO being the default setting


josevalim:
Hi @abc3 ! Just to be double sure, the `:lifo` value is not used anywhere yet, right?

Also, may I ask the use case for this feature? I am struggling to come up with cases where this won't cause massive latency to some requests. :)

abc3:
Do you mean why we switched to FIFO? LIFO is the [default strategy in poolboy](https://github.com/devinus/poolboy/blob/9212a8770edb149ee7ca0bca353855e215f7cba5/src/poolboy.erl#L47C35-L47C35), and we wanted to change it to FIFO because we believe that would be more intuitive

josevalim (commented):
LGTM!

chasers:
Should have used NimblePool ;)

chasers (approved):

PR #175, opened 2023-10-06
feat: support auth_query with md5

abc3:


lib/supavisor/protocol/server.ex
```diff
@@ -258,12 +265,15 @@ defmodule Supavisor.Protocol.Server do
     end
   end
 
-  @spec auth_request() :: iodata()
-  def auth_request() do
-    @auth_request
+  @spec auth_request(:scram | :md5) :: iodata
+  def auth_request(method \\ :scram, salt \\ nil) do
```
josevalim, line 269: I would split this in two functions, given the usage of default arguments and having one argument that only applies in one case:

    def md5_request(salt)
    def scram_request()

WDYT?

lib/supavisor/client_handler.ex
```diff
@@ -393,15 +393,19 @@ defmodule Supavisor.ClientHandler do
...
+  def parse_user_info(%{"user" => user}) do
+    case :binary.matches(user, ".") do
       [] ->
-        {username, nil}
+        {user, nil}
 
       matches ->
         {pos, _} = List.last(matches)
-        {name, "." <> external_id} = String.split_at(username, pos)
+        {name, "." <> external_id} = String.split_at(user, pos)
```
josevalim, line 408: split_at splits on length, :binary.matches  returns bytes, so if you have unicode characters in the user, this may fail. To be safe, you probably want something like this:

```elixir
{pos, 1}  = List.last(matches)
<<name::size(pos)-binary, ?., external_id::binary>> = user
```

PS: I didn't test it. :)
{pos, size} = List.last(matches)

josevalim (commented):
LGTM. :)

lib/supavisor/protocol/server.ex
abc3: sounds good 👍

lib/supavisor/client_handler.ex
abc3: good call, thanks!

delgado3d (approved):

PR #179, opened 2023-10-12
fix: cache for user query, server_name_indication, change log level

abc3:


delgado3d (approved):

lib/supavisor/db_handler.ex
```diff
@@ -225,7 +225,7 @@ defmodule Supavisor.DbHandler do
 
   def handle_event(:internal, :check_buffer, :idle, %{buffer: buff} = data) do
     if buff != [] do
-      Logger.warning("Buffer is not empty, try to send #{IO.iodata_length(buff)} bytes")
+      Logger.info("Buffer is not empty, try to send #{IO.iodata_length(buff)} bytes")
```
chasers, line 228: I think our log level is info on the server.

```suggestion
      Logger.debug("Buffer is not empty, try to send #{IO.iodata_length(buff)} bytes")
```

Until we get the dynamic log level setting?

lib/supavisor/db_handler.ex
```diff
@@ -252,7 +252,7 @@ defmodule Supavisor.DbHandler do
   end
 
   def handle_event({:call, {pid, _} = from}, {:db_call, bin}, state, %{buffer: buff} = data) do
-    Logger.warning(
+    Logger.info(
```
chasers, line 255:
```suggestion
    Logger.debug(
```

chasers (approved):
Approved pending passing actions

Issue #171, opened 2023-09-29
Provide a Docker Image?


mayel:
Hello, I was wondering if there's any plan/possibility to provide a supavisor docker image? This would make automated deploys easier.

ivanminutillo:
that would be very nice indeed!

chasers:
🎉🥳 🎊 🍾

https://hub.docker.com/r/supabase/supavisor/tags

PR #180, opened 2023-10-13
feat: add gh workflow

J0:

Aims to address #171 and also support Supabase Localhost.

Secrets need to be set before we can set this to ready

https://supabase.slack.com/archives/C04JR9DBNQL/p1697128718400279


.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,163 @@
...
+      - name: Merge multi-arch manifests for latest output
+        run: |
+          docker buildx imagetools create -t supabase/supavisor:latest \
+          supabase/supavisor@${{ needs.docker_x86_release.outputs.image_digest }} \
+          supabase/supavisor@${{ needs.docker_arm_release.outputs.image_digest }}
+
+      - name: configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.PROD_AWS_ROLE }}
```
J0, line 128: Staging AWS Role

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,163 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - id: semantic
+        uses: cycjimmy/semantic-release-action@v3
+        with:
+          semantic_version: 18
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_PROJECT_ACTION }}
+
+  docker_x86_release:
```
sweatybridge, line 21: https://github.com/supabase/postgres/blob/develop/.github/workflows/dockerhub-release.yml

Can we follow this new example using build matrix? It reduces quite a bit of duplication.

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,163 @@
+name: Build Production
+on:
+  push:
+    branches:
+      - "main"
+jobs:
+  release:
```
sweatybridge, line 7: Also the release step should be reading the version tag from file right? Since we don't use semantic release in supavisor repo

.github/workflows/publish_docker.yml
```diff
@@ -0,0 +1,98 @@
...
+          SLACK_WEBHOOK: ${{ secrets.SLACK_NOTIFICATIONS_WEBHOOK }}
+          SLACK_USERNAME: "gha-failures-notifier"
+          SLACK_COLOR: "danger"
+          SLACK_MESSAGE: "Building Postgres image failed"
+          SLACK_FOOTER: ""
+
+  publish:
+    needs: [settings, merge_manifest]
+    # Call workflow explicitly because events from actions cannot trigger more actions
+    uses: ./.github/workflows/mirror.yml
```
sweatybridge, line 95: remember to also add the mirror job to here https://github.com/supabase/postgres/blob/develop/.github/workflows/mirror.yml

J0: Thanks for flagging! Getting a bit late here, will circle back tmrw

chasers (approved):

PR #188, opened 2023-10-17
feat: add native mode

abc3:
This PR adds a new mode called "native". Its main responsibility is to retrieve the host and port information for a tenant from the connection string, establish a direct connection to the tenant's database, and then transfer the data without any modifications.


josevalim (commented):
Beautifully done, great use of proc_lib+enter_loop to handoff the socket/connection.

abc3:
@josevalim, thanks! Having reviews by Dashbit significantly raises the skills :relieved:

chasers (approved):

PR #194, opened 2023-10-20
feat: handle cancelling requests

abc3:
This PR adds support for query cancellation, allowing clients to terminate long-running queries. Additionally, this PR includes the forwarding of the db name from the client connection to pool initialization.

Links for context:
- [Overview of Internals](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-CANCELING-REQUESTS)
- [BackendKeyData Format](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-BACKENDKEYDATA)
- [CancelRequest Format](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-CANCELREQUEST)


lib/supavisor/db_handler.ex
```diff
@@ -110,6 +110,13 @@ defmodule Supavisor.DbHandler do
         %{tag: :ready_for_query, payload: db_state}, {ps, _} ->
           {ps, db_state}
 
+        %{tag: :backend_key_data, payload: payload}, acc ->
+          key = {data.tenant, self()}
```
josevalim, line 114: You may not need to have `self()` as part of the key. Instead you can just use `data.tenant` and use `Registry.values` to look up the values for a given key-pid pair.

lib/supavisor/protocol/server.ex
```diff
@@ -360,18 +361,13 @@ defmodule Supavisor.Protocol.Server do
...
   end
 
-  @spec backend_key_data() :: iodata()
+  @spec backend_key_data() :: {iodata(), binary}
   def backend_key_data() do
-    procid = System.unique_integer([:positive, :monotonic])
-    secret = Enum.random(0..9_999_999_999)
-    payload = <<procid::integer-32, secret::integer-32>>
+    pid = System.unique_integer([:positive, :monotonic])
+    key = Enum.random(0..9_999_999_999)
```
josevalim, line 367: This may be a better generator:

```suggestion
    key = :crypto.strong_rand_bytes(4)
```

josevalim:
The linked Postgres docs are helpful but I am struggling a bit to understand the BackendKeyData bits. When you connect, do you send the BackendKeyData to Postgres or do you receive it from them?

abc3:
@josevalim Sorry for the lack of explanation. 

DB clients receive BackendKeyData, which identifies a unique connection between the client and the database. We need to store these keys which are [received by the pool processes](https://github.com/supabase/supavisor/blob/91366f39a9d3b28c247f12fd6ed1c87ba455340d/lib/supavisor/db_handler.ex#L114-L116) (db_handler). Additionally, we need to [send our unique keys](https://github.com/supabase/supavisor/blob/91366f39a9d3b28c247f12fd6ed1c87ba455340d/lib/supavisor/client_handler.ex#L251) to clients connected to Supavisor. When clients request a cancel query, we [send this event to the client_handler](https://github.com/supabase/supavisor/blob/91366f39a9d3b28c247f12fd6ed1c87ba455340d/lib/supavisor/handler_helpers.ex#L100-L106). The client_handler then [sends it to the tenant's database using the key from the associated pool process](https://github.com/supabase/supavisor/blob/91366f39a9d3b28c247f12fd6ed1c87ba455340d/lib/supavisor/client_handler.ex#L82-L97)

josevalim:
I see. We received it from the database but, because it is a proxy, we need to generate ours too, but so far the ones we generate are not used for anything?

abc3:
They are used. When the client_handler sends BackendKeyData, it then [subscribes to this pid/key pair](https://github.com/supabase/supavisor/blob/29b6d7020191d326a5e01838d9c1c4759749e28f/lib/supavisor/handler_helpers.ex#L110). If the client sends a [cancellation request](https://github.com/supabase/supavisor/blob/29b6d7020191d326a5e01838d9c1c4759749e28f/lib/supavisor/client_handler.ex#L74-L79) in another connection, we will [receive it inside existing client_handler process](https://github.com/supabase/supavisor/blob/29b6d7020191d326a5e01838d9c1c4759749e28f/lib/supavisor/client_handler.ex#L82-L97)

josevalim:
Perfect, thank you! :shipit:

josevalim:
The complexity is not code-wise but the fact it is scattered across different modules. But it has to be scattered, because it comes from places with different responsibilities. Something that could help is to have some mermaid.js diagrams ([like we do here](https://hexdocs.pm/elixir/main/GenServer.html#module-client-server-apis)). They are supported by GitHub too. However, it is completely your decision if it is worth the time investment. Don't do it on my account. :)

PR #201, opened 2023-11-03
fix: docs enumerate tenant and user configs with auth_query

chasers:
- [x] lists `tenant` fields and descriptions
- [x] lists `user` fields and descriptions
- [x] new Configuration section
- [x] move Pool Modes subsection to Configuration section
- [x] adds Authentication section to detail `auth_query` usage vs `user` record usage


abc3 (approved):

docs/configuration/tenants.md
```diff
@@ -0,0 +1,41 @@
...
+
+`db_database` - the database of the Postgres instance
+
+`external_id` - an id used in an external system used to lookup a tenant
+
+`default_parameter_status` - the default initial connection parameters to use
+
+`ip_version` - the ip address type of the connection to the database server
+
+`upstream_ssl` - the ssl certificate to use when connecting to the database server
```
abc3, line 21:
```suggestion
`upstream_ssl` - enforce an SSL connection on the tenant's database
```

docs/configuration/tenants.md
```diff
@@ -0,0 +1,41 @@
...
+
+`upstream_verify` - how to verify the ssl certificate
+
+`upstream_tls_ca` - the ca certificate to use when connecting to the database server
+
+`enforce_ssl` - enforce an SSL connection on client connections
+
+`require_user` - require at least one `user` is created for a tenant
```
abc3, line 29: Required stored `users` with a pair of a username and password. As an opposite, only one user can be stored, through which the auth_query will be executed

abc3, line 29:
```suggestion
`require_user` - Required stored users with a pair of a username and password. As an opposite, only one user can be stored, through which the auth_query will be executed
```

docs/configuration/tenants.md
```diff
@@ -0,0 +1,41 @@
...
+
+`auth_query` - the query to use when matching credential agains a client connection
+
+`default_pool_size` - the default size of the database pool
+
+`sni_hostname` - the hostname expected on an SSL client connection
```
abc3, line 35:
```suggestion
`sni_hostname` - can be used to match a connection to a specific tenant record
```

docs/connecting/authentication.md
```diff
@@ -0,0 +1,42 @@
+When a client connection is established Supavisor needs to verify the credentials of the connection.
+
+Credential verificiation is done either via `user` records or an `auth_query`.
+
+## Tenant User Record
+
+If no `auth_query` exists on the `tenant` record credentials will be looked up from a `user` and verified against the client connection string credentials.
```
abc3, line 7: if Supavisor does not find the specified user/alias in the users table, it will attempt to find a user with `is_manager: true` set to true and retrieve creds using the auth query

docs/connecting/authentication.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Authentication Query
+
+If an `auth_query` exists on the `tenant` this query is used to query the tenant database and return results with credentials.
+
+Create a function to return a username and password for a user:
```
abc3, line 13: perhaps, for a quick review, it's worth mentioning that it's possible to use just the following:

```sql
SELECT rolname, rolpassword FROM pg_authid WHERE rolname=$1
```

docs/connecting/authentication.md
chasers: "If the `user` in the client connection is not found for a `tenant` it will use the `user` where `is_manager` is `true` and the `auth_query` on the `tenant` to return matching credentials from the tenant database."

PR #205, opened 2023-11-10
fix: add docs for migrating from pgbouncer

chasers:
- [x] add docs for migrating from PgBouncer
- [x] discusses handling two connection pools connected at the same time


docs/migrating/pgbouncer.md
```diff
@@ -0,0 +1,52 @@
...
+
+Once we've got Supavisor started we can verify it's using the amount of connections we set for `default_pool_size`:
+
+```sql
+select
+  count(*) as count,
+  usename,
+  application_name
+from pg_stat_activity
+where application_name ilike '%supavisor%'
```
abc3, line 43:
```suggestion
where application_name ilike '%Supavisor%'
```

abc3: I guess ILIKE is case-sensitive, right?

abc3 (approved):

docs/migrating/pgbouncer.md
chasers: `ilike` is case insensitive but we'll take it.

PR #207, opened 2023-12-06
feat: handling of named prepared statements

abc3:
This PR introduces support for prepared named statements in transaction mode. 

[Supavisor attempts to decode all incoming queries](https://github.com/supabase/supavisor/blob/493c6deac161fa13780e54d8eeb8528ca9dd7965/lib/supavisor/client_handler.ex#L807) from connected users and retrieves the SQL statement by parsing the decoded queries through an [existing Rustler wrapper](https://github.com/supabase/supavisor/blob/feat/named_ps/native/pgparser/src/lib.rs) around `libpg_query`. When the parsed query contains `PrepareStmt` or `DeallocateStmt` (which requires removing an existing named statement), the `ClientHandler` [broadcasts](https://github.com/supabase/supavisor/blob/493c6deac161fa13780e54d8eeb8528ca9dd7965/lib/supavisor/client_handler.ex#L813-L825) it to all other `DbHandlers` in the pool.

In `DbHandler`, these queries [accumulate in the `anon_buffer` list](https://github.com/supabase/supavisor/blob/5a1a860ee7abf68bed61978bd4fa237d6129b453/lib/supavisor/db_handler.ex#L327) and will be applied when [`data.caller` is `nil`](https://github.com/supabase/supavisor/blob/5a1a860ee7abf68bed61978bd4fa237d6129b453/lib/supavisor/db_handler.ex#L248-L257). They will wait [until the current PID is unlinked](https://github.com/supabase/supavisor/blob/5a1a860ee7abf68bed61978bd4fa237d6129b453/lib/supavisor/db_handler.ex#L313-L317) in cases where the process is reserved by another client

Screenshot 2023-12-07 at 11 02 40

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -334,27 +334,7 @@ defmodule Supavisor.ClientHandler do
   def handle_event(:info, {proto, _, bin}, :idle, %{pool: pid} = data) when is_pid(pid) do
     ts = System.monotonic_time()
     db_pid = db_checkout(:both, :on_query, data)
-
-    with {:ok, payload} <- Client.get_payload(bin),
-         {:ok, statamets} <- Supavisor.PgParser.statements(payload),
-         true <- Enum.member?([["PrepareStmt"], ["DeallocateStmt"]], statamets) do
-      Logger.info("Handle prepared statement #{inspect(payload)}")
-      {_, pid} = db_pid
-
-      GenServer.call(data.pool, :get_all_workers)
-      |> Enum.each(fn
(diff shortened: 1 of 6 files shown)


josevalim:
Excellent description and diagram. I have only one question though: @abc3 does it mean the statement will be prepared once per db handler? Would it make sense to keep the statements in memory and, given you parse queries, when there is an attempt to execute the statement, you immediately prepare it before (which then stays prepared)?

josevalim (approved):
Code wise the changes look great to me :D

abc3:
@josevalim Thanks for the review and the excellent suggestion for improvement 🔥 
We will try to implement it in the next iteration

PR #220, opened 2023-12-11
fix: return proper error on undefined tenant

### LINKED ISSUE #208: Admin API error: retrieving info for non-existent tenant
# Bug report

## Describe the bug

During the retrieval of information about a non-existing tenant, the backend responds with a `500` error instead of the expected `404`.

## To Reproduce

Try to call GET /api/tenants/undef_tenant:
```bash
$ curl -I -X GET 'http://localhost:4000/api/tenants/undef_tenant' --header 'Authorization: Bearer some_secret'

HTTP/1.1 500 Internal Server Error
cache-control: max-age=0, private, must-revalidate
content-length: 83348
content-type: text/html; charset=utf-8
date: Thu, 07 Dec 2023 15:26:39 GMT
server: Cowboy
x-request-id: F56WJhzGDekfNCYAAAJi

```

## Expected behavior

Backend should respond with a "not found" error (404)

## Possible solution

This happens because the [not_found.json](https://github.com/supabase/supavisor/blob/7baeb984a6259f984fdaa47b693cec54b80502b7/lib/supavisor_web/controllers/tenant_controller.ex#L75) render is missing in `SupavisorWeb.TenantView`. To fix this, it's necessary to add it in [SupavisorWeb.TenantView](https://github.com/supabase/supavisor/blob/main/lib/supavisor_web/views/tenant_view.ex)

```Elixir
def render("not_found.json", _) do
  %{error: "not found"}
end
```

J0:

naive implementation to fix  #208 Probably should add a test.


abc3:
add a test would be nice :smiley:

abc3 (approved):
lgtm!

PR #223, opened 2023-12-14
feat: add heartbeat interval to client handler

abc3:
We encounter scenarios where clients don't properly close their TCP connections.

These undead connections are undetectable through normal means and only reveal themselves when trying to transmit data to their sockets. To mitigate this, this PR introduces periodical TCP messages akin to heartbeats. 

[The PostgreSQL wire protocol allows only `NoticeResponse` and `ParameterStatus` for asynchronous client messages](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-ASYNC), so we will use ParameterStatus, specifically through the `application_name` field, as a heartbeat message.

```elixir
<<?S, 31::32, "application_name", 0, "Supavisor", 0>>
```


chasers (approved):

chasers:
Should we have tests for these timeouts?

lib/supavisor/client_handler.ex
```diff
@@ -828,4 +819,20 @@ defmodule Supavisor.ClientHandler do
...
+        Logger.debug("Call timeout #{inspect(t)}")
+        [timeout_check(:heartbeat_check, v) | acc]
+
+      {:idle_timeout, v} = t, acc when v > 0 ->
+        Logger.debug("Call timeout #{inspect(t)}")
+        [timeout_check(:idle_terminate, v) | acc]
+
+      _, acc ->
+        acc
+    end)
```
josevalim, line 836: Just a tiny suggestion that you can write code that has to filter and map as a `flat_map`. :)

```suggestion
    Enum.flat_map(data, fn
      {:heartbeat_interval, v} = t when v > 0 ->
        Logger.debug("Call timeout #{inspect(t)}")
        [timeout_check(:heartbeat_check, v)]

      {:idle_timeout, v} = t when v > 0 ->
        Logger.debug("Call timeout #{inspect(t)}")
        [timeout_check(:idle_terminate, v)]

      _ ->
        []
    end)
```

But if you prefer reduce, feel free to stick with it!

josevalim (commented):
LGTM!

abc3:
@josevalim Thanks for the review! Btw, have you encountered a similar issue? Are there any other methods to check if a TCP socket is alive, apart from sending data to it?

josevalim:
Ah, good question. There is a keepalive option in `gen_tcp` that will emit a `etimedout` if the keep alive fails. You could give it a try? It should definitely be less overhead than the current implementation. I incorrectly assumed that for some reason the TCP keep alive is not enough. :)

abc3:
Keepalive is enabled, but it doesn't help. I suppose the interval is defined at the OS level, and it is set to 2 hours by default? https://github.com/supabase/supavisor/blob/770a738da35e8dabf3c7d46548cdca630550fa28/lib/supavisor/application.ex#L43

josevalim:
Yup. Then the only other option I can think is doing it at the application level, as in this PR.

abc3:
Gotcha, thanks for the confirmation 🙏

Issue #236, opened 2023-12-23
Support network restrictions


chasers:
I should be able to add CIDR ranges to an `allow_list` field on the `tenant` and only allow connections from those ranges.

chasers:
[Closed](https://github.com/supabase/supavisor/pull/240)

PR #238, opened 2023-12-23
feat: add cidr ip allowlist

J0:

Aims to address #236. Bit late here so will probably close out once I wake up

FLUPs
- [ ] Check if we can use a custom Ecto type to leverage on the Postgres `inet` type
- [ ] Implement the IP check (check if it's in correct locaiton)
- [ ] Write a test (or maybe a few)


J0:
Superseded by #240  let’s continue discussion there

PR #240, opened 2023-12-28
feat: check ip against allow list

chasers:
Screenshot 2023-12-30 at 8 56 43 AM

- [x] Respond with Postgres error if client address is not in `tenant` `allow_list`
- [x] Include field in OpenAPI spec
- [x] Default allow_list to `["0.0.0.0/0", "::/0"]`
- [x] allow_list is non-nullable
- [x] Add allow list field to `tenant`
- [x] Function docs
- [x] Function doc tests
- [x] Update docs
- [x] Controller tests
- [x] Validate cidrs on `tenant` create and update
- [x] Check `allow_list` in `NativeHandler`
- [x] Should be automatically handled in `WsProxy` because it is just forwarding data to localhost `ClientHandler`
- [x] handle `:ssl` sockets

## Future work
- [ ] fix all calls to `Telem.client_join` in ClientHandler


lib/supavisor_web/open_api_schemas.ex
```diff
@@ -150,7 +152,12 @@ defmodule SupavisorWeb.OpenApiSchemas do
             },
             users: %Schema{type: :array, items: User},
             inserted_at: %Schema{type: :string, format: :date_time, readOnly: true},
-            updated_at: %Schema{type: :string, format: :date_time, readOnly: true}
+            updated_at: %Schema{type: :string, format: :date_time, readOnly: true},
+            allow_list: %Schema{
```
darora, line 156: Note that it wouldn't surprise me if in the future we want to add a block list as well. No change is needed right now, if you're happy with the current schema evolving from its current state to that.

lib/supavisor_web/views/tenant_view.ex
```diff
@@ -12,26 +12,7 @@ defmodule SupavisorWeb.TenantView do
...
-      auth_query: tenant.auth_query,
-      sni_hostname: tenant.sni_hostname,
-      default_pool_size: tenant.default_pool_size,
-      default_max_clients: tenant.default_max_clients,
-      client_idle_timeout: tenant.client_idle_timeout,
-      client_heartbeat_interval: tenant.client_heartbeat_interval,
-      default_pool_strategy: tenant.default_pool_strategy,
-      users: render_many(tenant.users, UserView, "user.json")
-    }
+    %{tenant | users: render_many(tenant.users, UserView, "user.json")}
```
abc3, line 15: do not print the certificate
```suggestion
    tenant = Map.drop(tenant, [:upstream_tls_ca])
    %{tenant | users: render_many(tenant.users, UserView, "user.json")}
```

lib/supavisor/handler_helpers.ex
```diff
@@ -126,4 +126,67 @@ defmodule Supavisor.HandlerHelpers do
...
+    []
+  end
+
+  @spec addr_from_port(port()) ::
+          {:ok,
+           {byte(), byte(), byte(), byte()}
+           | {char(), char(), char(), char(), char(), char(), char(), char()}}
+          | :error
+  def addr_from_port(port) do
+    case :inet.peername(port) do
```
abc3, line 175: This won't work for SSL connections. For a secured connection, you need to call [:ssl.peername/1](https://www.erlang.org/doc/man/ssl#peername-1). Therefore, you need to pass all `Supavisor.sock()` and then choose the right module: `ssl` or `inet`, based on the transport type (`ssl` or `gen_tcp`). Here are some similar functions https://github.com/supabase/supavisor/blob/9186398e5e3ff0e93254a797f2f14dceb207de99/lib/supavisor/handler_helpers.ex#L20-L33

lib/supavisor/handler_helpers.ex
```diff
@@ -126,4 +126,67 @@ defmodule Supavisor.HandlerHelpers do
...
+    []
+
+  """
+
+  @spec filter_cidrs(
+          list(),
+          {byte(), byte(), byte(), byte()}
+          | {char(), char(), char(), char(), char(), char(), char(), char()}
+          | any()
+        ) :: list()
```
abc3, line 157:
```suggestion
  @spec filter_cidrs(list(), :inet.ip_address() | any()) :: list()
```

lib/supavisor_web/views/tenant_view.ex
chasers: This [is happening](https://github.com/supabase/supavisor/blob/6754ce5cb9a648343c075d95ae1aa0d4d7e6cb31/lib/supavisor/tenants/tenant.ex#L13) in the `Tenant` now.

lib/supavisor/monitoring/telem.ex
```diff
@@ -49,7 +49,7 @@ defmodule Supavisor.Monitoring.Telem do
   end
 
   @spec client_join(:ok | :fail, S.id()) :: :ok
-  def client_join(status, {tenant, user, mode}) do
+  def client_join(status, {{_, tenant}, user, mode}) do
```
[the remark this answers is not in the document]
chasers: Well, still need to fix the rest of the calls to this. There is also the case where the `id` can be `nil`.

lib/supavisor_web/open_api_schemas.ex
chasers: That'd be fine. I like separate `allow_list` and `deny_list` fields I think anyways.

lib/supavisor/client_handler.ex
```diff
@@ -173,28 +173,41 @@ defmodule Supavisor.ClientHandler do
 
         Registry.register(Supavisor.Registry.TenantClients, id, [])
 
-        if info.tenant.enforce_ssl and !data.ssl do
-          Logger.error("Tenant is not allowed to connect without SSL, user #{user}")
+        {_type, port} = sock
+        {:ok, addr} = HH.addr_from_port(port)
```
abc3, line 177: the connection will crash if the function returns a non {:ok, _} response

abc3 (commented):
Looks awesome, just need to update `HandlerHelpers.addr_from_port/1`

lib/supavisor/handler_helpers.ex
chasers: Fixed!

lib/supavisor/client_handler.ex
chasers: This should be okay yeah? 

It will fail closed, users just won't get a nice error message (should see it in the logs still).

I didn't want to rewrite too much of the client handler.

Probably should move to a `with` clause here as we're doing more than 3 checks in nested conditions.

abc3 (approved):
🔥

priv/repo/migrations/20231229010413_add_tenant_allow_list.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Supavisor.Repo.Migrations.AddTenantAllowList do
+  use Ecto.Migration
+
+  def change do
+    alter table("tenants", prefix: "_supavisor") do
+      add(:allow_list, {:array, :string}, null: false, default: ["0.0.0.0/0", "::/0"])
```
philss, line 6: It's fine the way it is, but just for completeness, PG has these [Network Address Types](https://www.postgresql.org/docs/current/datatype-net-types.html) that may be helpful to keep the data integrity at the DB level :)

lib/supavisor/handler_helpers.ex
```diff
@@ -126,4 +126,68 @@ defmodule Supavisor.HandlerHelpers do
     :ok = sock_send(sock, msg)
     :ok = sock_close(sock)
   end
+
+  @doc """
+  Takes an allow list of CIDR ranges and filtres them for ranges which contain the address
```
philss, line 131: I believe this is a typo: `filtres` => `filters`

lib/supavisor/tenants/tenant.ex
```diff
@@ -71,9 +75,60 @@ defmodule Supavisor.Tenants.Tenant do
...
+
+  defp valid_range?(range) do
+    try do
+      InetCidr.parse(range)
+      true
+    rescue
+      _e ->
+        false
+    end
+  end
```
philss, line 133: It's fine the way it is, it's also possible to use the shorter of `try/rescue` in a function:

```elixir
defp valid_range?(range) do
  InetCidr.parse(range)
  true
rescue
  _e ->
    false
end
```

philss (commented):
LGTM! I added some minor things :)

Happy new year! :D

priv/repo/migrations/20231229010413_add_tenant_allow_list.exs
chasers: Ah yeah nice! Will make another pass with this in a bit.

Issue #234, opened 2023-12-21
Supavisor forces request to "postgres" database even when other database is specified


gitaugakwa:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When attempting to connect to a database that is not named "postgres", supavisor forces the connection to the "postgres" database. This was previously mentioned in the issue https://github.com/supabase/supavisor/issues/187 but has come up again. I had the following discussion on [Discord](https://discord.com/channels/839993398554656828/1187460410782257154).

## To Reproduce

Create a connection to a database other than "postgres" through Supavisor.
Make a query through said connection.
Query would be performed on "postgres" database.

## Expected behavior

Query to be performed on the database stated other than "postgres".

## Screenshots

![Image of Environment Variables used to create connection](https://github.com/supabase/supavisor/assets/50299997/745240e0-67d3-4787-80b3-d18d31b5f2ca)

## System information

- OS: Windows
- Version of supavisor: Version used on supabase.com as of 21/12/2023
- Version of Node.js: v18.17.1

## Additional context

[See the Discord thread linked](https://discord.com/channels/839993398554656828/1187460410782257154)

yharaskrik:
I am also seeing this. We have different databases that we manage via Prisma (that won't be public and we don't want to manage RLS rules on) and would love to be able to use pooling!

chasers:
Got it ... we will fix.

abc3:
Hey, just wanted to inform you that we've recently deployed a fix. We believe the problem should now be resolved. Please feel free to check and confirm whether everything is functioning as expected on your end

PR #254, opened 2024-01-13
fix: enable connection to custom database name other than 'postgres'

w3b6x9:

Bug fix

I can only connect to the `postgres` database when connecting via Supavisor pooler.

## What is the new behavior?

I can connect to another database other than `postgres` database.

## Additional context

This addresses the GH issue [Supavisor forces request to "postgres" database even when other database is specified #234](https://github.com/supabase/supavisor/issues/234).

While this solution enables connecting to other database, that's not `postgres` database, it presents other problems.

Here are some things that should be tackled alongside this change for completeness:

1. As Supavisor currently stands, this change enables listening to other database when the tenant has no existing local pool. However, if tenant wants to listen to other database then `postgres` database then it will default to other database. Whichever database a tenant connects to first will be the default one.
~2. Supavisor will continuously attempt to connect to a database even when that database doesn't exist. However, the client doesn't get the proper error feedback that database doesn't exist. See screen recording:~

Spun this out into its own issue: https://github.com/supabase/supavisor/issues/257

https://github.com/supabase/supavisor/assets/5532241/6b07f139-66be-4d7b-965a-bb769df9c037


docs/development/setup.md
```diff
@@ -26,13 +26,14 @@ curl -X PUT \
...
-    "default_pool_size: 15,
+    "default_pool_size": 15,
     "users": [
       {
         "db_user": "postgres",
         "db_password": "postgres",
         "mode_type": "transaction",
-        "pool_checkout_timeout": 100
+        "pool_checkout_timeout": 100,
+        "pool_size": 10
```
w3b6x9, line 36: I was getting an error without this.

lib/supavisor_web/ws_proxy.ex
```diff
@@ -61,6 +61,6 @@ defmodule SupavisorWeb.WsProxy do
   @spec connect_local() :: {:ok, port()} | {:error, term()}
   defp connect_local() do
     proxy_port = Application.fetch_env!(:supavisor, :proxy_port_transaction)
-    :gen_tcp.connect('localhost', proxy_port, [:binary, packet: :raw, active: true])
+    :gen_tcp.connect(~c"localhost", proxy_port, [:binary, packet: :raw, active: true])
```
abc3, line 64: I think we don't need sigils here, single quotes are a pretty native method for creating charlists

test/integration/proxy_test.exs
```diff
@@ -170,10 +170,10 @@ defmodule Supavisor.Integration.ProxyTest do
     assert :httpc.request(
              "http://localhost:#{Application.get_env(:supavisor, :proxy_port_transaction)}"
            ) ==
-             {:ok, {{'HTTP/1.1', 204, 'OK'}, [], []}}
+             {:ok, {{~c"HTTP/1.1", 204, ~c"OK"}, [], []}}
```
abc3, line 173: the same about ~c

abc3:
legend 🔥

lib/supavisor_web/ws_proxy.ex
w3b6x9: this was result of `mix format`

test/integration/proxy_test.exs
w3b6x9: this was result of `mix format`

lib/supavisor_web/ws_proxy.ex
w3b6x9: reverted back

test/integration/proxy_test.exs
w3b6x9: reverted back

abc3 (approved):

PR #280, opened 2024-01-23
chore: set max heap size for handlers

abc3:
To prevent out of memory issues, this PR sets the `max_heap_size` for `DbHandler` and `ClientHandler` to 150 MB


josevalim (commented):
LGTM! I generally prefer to pass these values as options on proc_lib.spawn_opt and start_link, but setting it via a flag should work just as well!

Issue #209, opened 2023-12-07
Option to set a log level on a client connection


abc3:
# Enhancement

## Describe the problem

Debugging client issues can sometimes be challenging. It would be beneficial to have the ability to set a specific log level for each individual client connection.

## Possible implementation

This could be achieved by passing a log level in the `options` parameter:

```bash
$ psql 'postgres://postgres.tenant@localhost:6543/postgres?options=log_level%3dinfo'
```

Then, this value will be available in the [`startup`](https://github.com/supabase/supavisor/blob/7baeb984a6259f984fdaa47b693cec54b80502b7/lib/supavisor/client_handler.ex#L135-L136) message:

```elixir
%{
  len: 110,
  payload: %{
    "application_name" => "psql",
    "client_encoding" => "UTF8",
    "database" => "postgres",
    "options" => %{"log_level" => "info"},
    "user" => "postgres.tenant"
  },
  tag: :startup
}
```

After that, to change the log level of the current process, you need to invoke:

```elixir
Logger.put_process_level(self(), log_level)
```
*In relation to the existing code, the value is accessible in the [`hello`](https://github.com/supabase/supavisor/blob/7baeb984a6259f984fdaa47b693cec54b80502b7/lib/supavisor/client_handler.ex#L135C22-L135C22) variable. You can try 
Logger.put_process_level(self(), String.to_existing_atom(hello.payload["options"]["loglevel"]))*

Moreover, we could extend the [Supavisor.start/3](https://github.com/supabase/supavisor/blob/7baeb984a6259f984fdaa47b693cec54b80502b7/lib/supavisor/client_handler.ex#L228) function to accept a log level as a parameter and start all processes related to a specific tenant with this log level.

PR #286, opened 2024-01-25
feat: setting log level per tenant via options

abc3:
implements: https://github.com/supabase/supavisor/issues/209


lib/supavisor.ex
```diff
@@ -16,19 +16,19 @@ defmodule Supavisor do
 
   @registry Supavisor.Registry.Tenants
 
-  @spec start_dist(id, secrets, Node.t() | boolean()) ::
+  @spec start_dist(id, secrets, atom(), Node.t() | boolean()) ::
           {:ok, pid()} | {:error, any()}
-  def start_dist(id, secrets, force_node \\ false) do
+  def start_dist(id, secrets, log_level \\ nil, force_node \\ false) do
```
wojtekmach, line 21: function with multiple defaults are sometimes confusing to use and are hard to grow in the future. It's fine to leave this as is but you can also bite the bullet right now and convert to options.

```suggestion
  def start_dist(id, secrets, options \\ []) do
    options = Keyword.validate!(options, log_level: nil, force_node: false)
    log_level = Keyword.fetch!(options, :log_level)
    force_node = Keyword.fetch!(options, :force_node)
```

wojtekmach (commented):
Looks good to me!

filipecabaco (approved):

PR #291, opened 2024-01-30
feat: reset the check timer after the client's disconnection

abc3:
This PR resets the check_subscribers timer after each disconnection


josevalim:
@abc3 can you expand it a bit more on what you are seeing? Could you rather have concurrent updates on the table? If concurrent updates are not expected, what about setting the table as `:protected`, to guarantee you are not running into concurrency issues?

abc3:
@josevalim, sometimes, we see that a client is trying to subscribe to a recently stopped manager. I think it's because the current check_timeout simply occurs every 2 minutes and can catch a short moment when a tenant doesn't have any subscribers, leading to the connection being closed. Ets with subscribers can only be updated from a single GenServer. I just want to be sure that the ets size always shows the actual number of records in the table

josevalim:
If the size is read within the same process doing the writes, then it should not be out of sync, yeah. If you only write from a process, I recommend changing the table from :public to :protected. :)

abc3:
@josevalim I'll change it to protected, thanks!

PR #304, opened 2024-02-14
feat: recheck auth secrets on connection error

abc3:
In this PR, when a user receives a "wrong password" error, we will check the authentication secrets. If they are different from what is already stored, we will update the cache for these secrets and stop the connections tree.

Also included in this PR:

- the default pool size for the internal database has been set to 50
- changed the log level to debug when an unlinked DbHandler receives a database response
- the cache for secrets has been set to 24 hours
- the cache for tenants has been set to 24 hours (it is dropped when a tenant is updated via the API)
- logs erlang system monitor events


wojtekmach (commented):
Looks good to me. Consider adding some regression tests.

kiwicopple (approved):

PR #309, opened 2024-02-19
feat: clear tenant's cache on update or delete

abc3:
In this PR, we have introduced functionality to clear the entire internal cache for a specific tenant when an update or delete operation is performed through an API call. Additionally, we have disabled cache clearing when the pool is terminated.


lib/supavisor.ex
```diff
@@ -160,6 +160,34 @@ defmodule Supavisor do
     end)
   end
 
+  @spec del_all_cache(String.t()) :: [map()]
+  def del_all_cache(tenant) do
+    Logger.info("Deleting all cache for tenant #{tenant}")
+    {:ok, keys} = Cachex.keys(Supavisor.Cache)
```
chasers, line 166: I've had issues dumping all the keys of a cache frequently. This may not happen frequently and it may not be a terribly large key space. Mentioning just in case.

You could instead use a Cachex.Query to filter the keys for a tenant.

[Here's a query example](https://github.com/Logflare/logflare/blob/22a79f98a0d5cd6dd439310d53a5565f61497a25/lib/logflare/context_cache.ex#L80) I did with Logflare recently.

chasers:
Is there a reason you're doing those inspects everywhere? Seems odd idk.

lib/supavisor.ex
abc3: ohh, cool

chasers (approved):

lib/supavisor.ex
abc3: I think it should be fine with only `Cachex.stream!()`

abc3:
I use inspect() to make valid map's key/value pairs, and then it will be possible to safely encode them to JSON.

philss (commented):
LGTM!

lib/supavisor/syn_handler.ex
```diff
@@ -16,7 +16,7 @@ defmodule Supavisor.SynHandler do
 
     # remove all Prometheus metrics for the specified tenant
     PromEx.remove_metrics(id)
-    Supavisor.del_all_cache(tenant, user)
+    # Supavisor.del_all_cache(tenant, user)
```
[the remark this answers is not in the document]
abc3: yep!

lib/supavisor.ex
```diff
@@ -160,6 +161,37 @@ defmodule Supavisor do
...
+      end
+    end)
+  end
+
+  @spec del_all_cache_dist(String.t(), pos_integer()) :: [map()]
+  def del_all_cache_dist(tenant, timeout \\ 15_000) do
+    Logger.info("Deleting all dist cache for tenant #{tenant}")
+
+    for node <- [node() | Node.list()] do
+      %{inspect(node) => :erpc.call(node, Supavisor, :del_all_cache, [tenant], timeout)}
```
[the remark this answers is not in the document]
abc3: I chose inspect() because it feels "safe" :), but yeah, I will replace it with to_string (even though it's possible to use just the node name as a key since it's an atom)

PR #316, opened 2024-03-05
chore: workflow for building soft upgrade tarball

abc3:


filipecabaco (approved):
Not fully sure how all of this works honestly but will leave the +1 to unblock if needed to test the pipeline

PR #318, opened 2024-03-07
fix: async calls for all ready_for_query messages

abc3:
The client handler and the DB handler communicate by sending messages to each other through a blocking handler call method (the first 1000 calls are async). This PR addresses issues that occur when the assigned DB connection sends a blocking ReadyForQuery message with either the transaction block indicator or a failed transaction block.

[ReadyForQuery description](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-READYFORQUERY)


chasers (approved):

lib/supavisor/db_handler.ex
```diff
@@ -542,10 +544,29 @@ defmodule Supavisor.DbHandler do
...
+            end
+
+          {:ready_for_query, indicator}
+
+        _ ->
+          :continue
+      end
+    else
+      :continue
+    end
```
josevalim, line 570: `<<_::binary-size(string_size - 6)>>` is always an error if size is not 6, so I think you can rewrite this to:

```suggestion
    case bin do
      <<90, 0, 0, 0, 5, status_indicator>> ->
          case status_indicator do
            ?I -> :idle
            ?T -> :transaction_block
            ?E -> :failed_transaction_block
            _ -> :continue # maybe this can be removed now?
          end

          {:ready_for_query, indicator}

      _ ->
        :continue
    end
```

WDYT?

josevalim: Ignore me.

josevalim, line 570: Ok, take 2. The other of `6 <= string_size` threw me for a spin. This should work as well:

```suggestion
    case bin do
      <<_::binary-size(string_size - 6), 90, 0, 0, 0, 5, status_indicator::binary>> ->
        indicator =
          case status_indicator do
            <<?I>> -> :idle
            <<?T>> -> :transaction_block
            <<?E>> -> :failed_transaction_block
            _ -> :continue
          end

        {:ready_for_query, indicator}

      _ ->
        :continue
    end
```

Although it is kind of hidden that the pattern matching fails if string_size is less than 6. It does help remove one level of nesting. The tests in this PR still pass. Your call. :)

josevalim (commented):
Looks good! Just one tiny comment :)

lib/supavisor/db_handler.ex
abc3: The main reason I added the if case is because it will end the function when the bin is smaller than 6 bytes. But I guess more often the bin is more than 6. Thanks for the suggestion!

```
Name                                ips        average  deviation         median         99th %
check_ready 3 bytes             28.53 M       35.06 ns   ±105.86%          42 ns          42 ns
check_ready_no_if 3 bytes       24.17 M       41.37 ns ±10422.75%          42 ns          42 ns
check_ready_no_if long          22.49 M       44.46 ns  ±9921.49%          42 ns          83 ns
check_ready_no_if short         22.29 M       44.86 ns ±10323.95%          42 ns          83 ns
check_ready short               22.03 M       45.40 ns ±10105.94%          42 ns          84 ns
check_ready long                21.86 M       45.75 ns ±10854.38%          42 ns          84 ns

Comparison:
check_ready 3 bytes             28.53 M
check_ready_no_if 3 bytes       24.17 M - 1.18x slower +6.31 ns
check_ready_no_if long          22.49 M - 1.27x slower +9.40 ns
check_ready_no_if short         22.29 M - 1.28x slower +9.80 ns
check_ready short               22.03 M - 1.30x slower +10.35 ns
check_ready long                21.86 M - 1.31x slower +10.69 ns
```

PR #321, opened 2024-03-09
fix: stop pool connection if credentials fail

abc3:


config/runtime.exs
```diff
@@ -154,7 +154,8 @@ if config_env() != :test do
     prom_poll_rate: System.get_env("PROM_POLL_RATE", "15000") |> String.to_integer(),
     global_upstream_ca: upstream_ca,
     global_downstream_cert: downstream_cert,
-    global_downstream_key: downstream_key
+    global_downstream_key: downstream_key,
+    reconnect_on_db_close: System.get_env("RECONNECT_ON_DB_CLOSE") == "true"
```
chasers, line 158: What's the point of the env var exactly?

abc3: to configure the behavior of db_handlers when Supervisor loses connection with the tenant's db. It allows to choose between two options: reconnecting to the database (the current behavior) or stopping the db_handler (which will become the default setting). If a large pool of db_handlers attempts to reconnect without valid credentials, the Supervisor IP will be banned

abc3: https://github.com/supabase/supavisor/blob/c6dc7d27531fdc4fbe5aecb6413e07b09ff02277/lib/supavisor/db_handler.ex#L417-L419

chasers: Why is it an env var if it's going to be the default is what I mean?

chasers:
I don't really see the reason to make this an env var but if you want to that is fine.

chasers (approved):

Issue #328, opened 2024-03-24
Docs should mention  VAULT_ENC_KEY size needs to be 32 chars


hgezim:
Spent a lot of time trying to find the cause of this.

hgezim:
Related https://github.com/supabase/supavisor/issues/211

J0:
Hey @hgezim,

I'll update the docs to clarify this - quick check in what context did you run into this error? (e.g.  while following `https://supabase.github.io/supavisor/development/setup/`  trying to deploy etc)

J0:
Updated docs to reflect this

Issue #327, opened 2024-03-22
Specify where to get auth token of user to create first tenant


hgezim:
# Improve documentation

## Link 

https://supabase.github.io/supavisor/development/setup/

## Describe the problem

I'm wrecking my brain trying to come up with that first user and how to get that token.

## Describe the improvement

I would like to know how that initial user is generate. I do not know how that's done.

## Additional context

I'm running supavisor using the docker-compose.yml provided in this repo. Many thanks for that.

J0:
Hey @hgezim,

Thanks for the query! Did you already try the token provided? I haven't tried in a while but I think it should work. Do you mind sharing the error if it doesn't? I'll take a look. I do know that I used a token with the role modified to `manager` a while back and I think that worked.

hgezim:
@J0 I'm getting this error when trying to call the PUT endpoint. It should be noted that `_supavisor` schema tables are not populated excepts for the migrations table:

```
> docker compose logs --follow supavisor
supavisor  | Setting RLIMIT_NOFILE to 100000
supavisor  | 05:38:16.143 [info] == Running 20230125140723 Supavisor.Repo.Migrations.CreateTenants.change/0 forward
supavisor  | 05:38:16.145 [info] create table _supavisor.tenants
supavisor  | 05:38:16.153 [info] create index _supavisor.tenants_external_id_index
supavisor  | 05:38:16.157 [info] == Migrated 20230125140723 in 0.0s
supavisor  | 05:38:16.180 [info] == Running 20230418151441 Supavisor.Repo.Migrations.CreateUsers.change/0 forward
supavisor  | 05:38:16.181 [info] create table _supavisor.users
supavisor  | 05:38:16.186 [info] create index _supavisor.users_db_user_alias_tenant_external_id_mode_type_index
supavisor  | 05:38:16.188 [info] == Migrated 20230418151441 in 0.0s
supavisor  | 05:38:16.190 [info] == Running 20230502101623 Supavisor.Repo.Migrations.AddTimeoutToUsers.up/0 forward
supavisor  | 05:38:16.190 [info] alter table _supavisor.users
supavisor  | 05:38:16.191 [info] == Migrated 20230502101623 in 0.0s
supavisor  | 05:38:16.192 [info] == Running 20230601125553 Supavisor.Repo.Migrations.AddTenantDefaultPS.up/0 forward
supavisor  | 05:38:16.192 [info] alter table _supavisor.tenants
[...]

PR #329, opened 2024-03-27
docs: update development docs

J0:

Add documentation that users can copy and paste to get started in addition in addition to the detailed command reference.

Also adds a note about `VAULT_ENC_KEY` to address #328. Would also help with #327  I guess


chasers (approved):
Nice!

Issue #206, opened 2023-11-24
Connection Pooler string does not honor options parameter with search_path


abdullahsych:
# Bug report

- [ x] I confirm this is a bug with Supabase, not with my own application. 
- [ x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Postgres connection strings support an `options` query parameter where one can add a specific `search_path` to be used for the connection:

```
POSTGRES_CONNECTION_STRING?options=-c%20search_path=custom_schema
```

This works if I use the direct Postgres connection string shown on the Supabase dashboard but does not work with the connection pooler string (Supavisor). Is this by design? I ran multiple tests and the search path in my connection only updates when I use the direct connection string.

## Expected behavior

The search path to be set as specified in the additional parameters passed to the connection pooler string.

## Additional context

I am using the connection pooler in Transaction mode.

mrkswrnr:
In addition, the supavisor seems to ignore the `SEARCH_PATH`, which is set via 
```sql
ALTER ROLE user in DATABASE database SET search_path = "user";
```
or
```sql
ALTER ROLE user SET search_path = "user";
```
For a direct connection, 
```sql
SELECT current_schema()
```
returns "user"; for a connection via the supavisor, the query returns no result.

This is a major problem if the user is to use a different schema by default and may even lack the rights to the default schema. As many applications in which you can configure a connection string do not offer the option of executing an initial command, there is no other option than to continue using the direct connection to the database to change the schema for a user.

Using the connection pooler in session mode or the connection via port 5432 (instead of 6543) does not seem to make any difference here.

l2dy:
Did https://github.com/supabase/supavisor/commit/163e4fa2269a3cf57ebf92b80d0a5c04435a4f44 address your issue?

olivernaaris:
I was also using latest version of Supabase with connection pooler in transaction mode and the custom schemas didn't work with the pooler, they however worked with direct connection.
Somewhat annoying that this issue has still not been fixed.
In the end I switched from Supabase connection pooler to pgcat, which at least works as intended.

drftrcld:
Leaving this here, since this worked for me:
As a workaround using the alternative syntax seems to fix the issue; e.g.:
```
      options: `--search_path=${schema}`,
```

mizoz:
[...]

PR #330, opened 2024-03-27
feat: support custom search_path

J0:

Intention to address #206. Need to look  a bit further though


J0:
Ah oops it's been a while. Will pick this up again soon.

PR #333, opened 2024-04-03
feat: encrypt secrets for internal use

abc3:
To avoid logging sensitive data (passwords, secrets, etc.), we keep it internally as anonymous functions `fn -> some_secret end`. However, during hot upgrades, old links to these funs can become invalid, so this PR introduces a new approach by turning sensitive data into binary.

The PR also includes the HotUpgrade module, which will replace existing encoded secrets, currently cached or residing within the `:poolboy` supervisor, with this new method


abc3:
@supabase/dashbit, if there is any way to continue keeping secrets in funs across upgrades, I would be happy to hear about it, as I haven’t found any safe method

lib/supavisor/helpers.ex
```diff
@@ -354,4 +354,41 @@ defmodule Supavisor.Helpers do
...
+    |> :erlang.term_to_binary()
+    |> Supavisor.Vault.encrypt!()
+  end
+
+  # for backward compatibility, will be removed in the future
+  @spec decode_secret(fun()) :: term()
+  def decode_secret(fun) when is_function(fun) do
+    Logger.warning(
+      "Using deprecated function `decode_secret/1` #{inspect(Process.info(self(), :current_stacktrace), pretty: true)}"
+    )
```
philss, line 383: Emitting the deprecation warning is fine, but have you considered using the `@deprecated` module tag - https://hexdocs.pm/elixir/Module.html#module-deprecated-since-v1-6-0 - with a different function name? This would make easier to spot the locations you are still using this, and would make easier to migrate. Of course is totally fine to keep as it is :)
WDYT?

philss (commented):
Hi @abc3 :wave: LGTM!

> if there is any way to continue keeping secrets in funs across upgrades, I would be happy to hear about it, as I haven’t found any safe method

Not that I know. I will keep investigating this, but I can only think of doing _serialization/deserialization_ of the function term, which is similar to what you did here. But yeah, it wouldn't be safe without the encryption/decryption.

chasers (approved):

josevalim:
Other approaches to consider:

1. Put those secrets into a struct and derive the implementation of the struct to not include the field:

```elixir
@derive {Inspect, only: []}
defstruct :secret
```

2. Store them in the process dictionary (they will be plain but it should not show up in logs)

lib/supavisor/hot_upgrade.ex
```diff
@@ -116,8 +119,8 @@ defmodule Supavisor.HotUpgrade do
 
         args =
           Map.update!(args, :auth, fn auth ->
-            Map.put(auth, :password, enc(auth.password.()))
-            |> Map.put(:secrets, enc(auth.secrets.()))
+            Map.put(auth, :password, enc(auth.password))
+            |> Map.put(:secrets, enc(auth.secrets))
```
filipecabaco, line 123:
```suggestion
            auth
            |> Map.put(:password, enc(auth.password))
            |> Map.put(:secrets, enc(auth.secrets))
```

filipecabaco (approved):

PR #342, opened 2024-05-27
feat: use simple_connection for auth_query

abc3:
The PID of the Postgrex connection can be lost when ClientHandler terminates in the middle of getting secrets via auth_query. As a result, we may get zombie connections, which can consume all allowed connections to the tenant's database. With this PR, auth_query will be done with SimpleConnection, where the process crashes automatically if the caller (ClientHandler) goes down.


lib/supavisor/helpers.ex
```diff
@@ -88,7 +88,9 @@ defmodule Supavisor.Helpers do
   @spec get_user_secret(pid(), String.t(), String.t()) :: {:ok, map()} | {:error, String.t()}
   def get_user_secret(conn, auth_query, user) do
     try do
-      Postgrex.query!(conn, auth_query, [user])
+      # Postgrex.query!(conn, auth_query, [user])
```
filipecabaco, line 91:
```suggestion
```

lib/supavisor/helpers.ex
```diff
@@ -88,7 +88,9 @@ defmodule Supavisor.Helpers do
...
+      auth_query = String.replace(auth_query, "$1", "'#{user}'")
```
filipecabaco, line 92: isn't this a potentially danger for sql injection?

abc3: I'll have to check, but good call!

josevalim (commented):
LGTM! :)

lib/supavisor/helpers.ex
```diff
@@ -88,7 +88,11 @@ defmodule Supavisor.Helpers do
   @spec get_user_secret(pid(), String.t(), String.t()) :: {:ok, map()} | {:error, String.t()}
   def get_user_secret(conn, auth_query, user) do
     try do
-      Postgrex.query!(conn, auth_query, [user])
+      # sanitize the user input by removing all characters that are not alphanumeric or underscores
+      user = String.replace(user, ~r/[^a-zA-Z0-9_]/, "")
```
abc3, line 92: @filipecabaco wdyt?

test/integration/single_connection_test.exs
```diff
@@ -0,0 +1,36 @@
...
+      assert %Postgrex.Result{rows: [["1"]]} =
+               Postgrex.SimpleConnection.call(pid, {:query, "SELECT 1"})
+    end)
+
+    :timer.sleep(250)
+
+    # check that the connection dies after the caller dies
+    assert Enum.filter(Process.list(), fn pid ->
+             Process.info(pid)[:dictionary][:auth_host] == db_conf[:hostname]
+           end) == []
```
hauleth, line 34:
```suggestion
    assert Enum.all?(Process.list(), fn pid ->
             Process.info(pid)[:dictionary][:auth_host] == db_conf[:hostname]
           end)
```

lib/supavisor/helpers.ex
hauleth: I think that instead we should reject `user` if these characters are present instead of just pretending that these do not exists.

test/integration/single_connection_test.exs
josevalim: For what is worth, using filter for this test has a tiny benefit that it shows the left side with the failed PIDs in exception reports. :)

hauleth: True. That indeed may be useful.

lib/single_connection.ex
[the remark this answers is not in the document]
abc3: oops, missed it

abc3:
@hauleth I already have a check for invalid characters during connection, but it only checks for slashes. I added this sanitization as an additional guard. It might be worth applying the same rule (`a-zA-Z0-9_`) during connection as well

https://github.com/supabase/supavisor/blob/61b629dc1995a6c1245c6d359c7a9f647181421b/lib/supavisor/client_handler.ex#L157-L164

```bash
$ psql postgresql://p\"ostgres.tenant:password@localhost:5432/postgres
psql: error: connection to server at "localhost" (::1), port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  Authentication error, reason: "Invalid characters in user or db_name"
```

lib/supavisor/helpers.ex
abc3: I already have a check for invalid characters during connection, but it only checks for slashes. I added this sanitization as an additional guard. It might be worth applying the same rule (`a-zA-Z0-9_`) during connection as well

https://github.com/supabase/supavisor/blob/61b629dc1995a6c1245c6d359c7a9f647181421b/lib/supavisor/client_handler.ex#L157-L164

```bash
$ psql postgresql://p\"ostgres.tenant:password@localhost:5432/postgres
psql: error: connection to server at "localhost" (::1), port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  Authentication error, reason: "Invalid characters in user or db_name"
```

Issue #343, opened 2024-05-28
Prepend application_name from client to application_name set by DbHandler only in session mode


chasers:
When looking at connections in Postgres I should be able to see the application_name set by clients if connected via session mode through Supavisor.

select usename, application_name from pg_stat_activity;

application_name should be `my_special_application_name via Supavisor`

kepler:
Any reason why this isn't currently possible? It would greatly help debugging connection issues.

spormeon:
i've been trying to pass this as well in the connection query string but I cant get it to work/ show up?  would also be good if we could pass other "metadata" as well for tracking debugging , such as zoneid, podid, regionid, as I have pods going up/ down like yoyo's adn would be good to see thse in logs etc?

mizoz:
Hi! I'm interested in contributing to Supavisor. This issue looks like a good fit — could you assign it to me?

Snehil-Shah:
@v0idpwn I think we can close this?

v0idpwn:
Let's close it when 2.10 is released

PR #345, opened 2024-05-29
docs: update Supavisor application_name query

w3b6x9:


docs/migrating/pgbouncer.md
```diff
@@ -40,7 +40,7 @@ select
   usename,
   application_name
 from pg_stat_activity
-where application_name ilike '%Supavisor%'
+where application_name = 'Supavisor'
```
w3b6x9, line 43: there's no point in doing `ilike` when we set it as `Supavisor`.

chasers: There is also a `Supavisor auth_query` one

docs/migrating/pgbouncer.md
```diff
@@ -40,7 +40,7 @@ select
   usename,
   application_name
 from pg_stat_activity
-where application_name ilike '%Supavisor%'
+where application_name in ('Supavisor', 'Supavisor auth_query')
```
w3b6x9, line 43: @chasers including the auth query one.

abc3 (changes requested):
conflicts with [#343](https://github.com/supabase/supavisor/issues/343)

docs/migrating/pgbouncer.md
chasers: But why? Now we have to remember to add any here if we update them.

PR #352, opened 2024-06-05
fix: sanitize tokens for the blacklist

abc3:


lib/supavisor_web/router.ex
```diff
@@ -89,6 +89,7 @@ defmodule SupavisorWeb.Router do
     blocklist = Application.fetch_env!(:supavisor, blocklist_key)
 
     with ["Bearer " <> token] <- get_req_header(conn, "authorization"),
+         token <- Regex.replace(~r/\s|\n/, URI.decode(token), ""),
```
filipecabaco, line 92: why not 
```
blocklist =
      Application.get_env(:realtime, blocklist_key, [])
      |> Enum.map(&String.trim/1)
      |> Enum.map(&String.downcase/1)

```

abc3: I repeated the same approach as in JwtAuthToken to ensure we manipulate the same string

filipecabaco: but why URI decode it? really curious about it.

abc3: you can hide some symbols via HTML characters, which token sanitization can skip but the JWT library can decode

filipecabaco: TIL 👀 will copy pasta it into realtime 🍝

filipecabaco: should we do the same to the blocklist tokens?

abc3: I guess not, because we can check that they are without spaces or new lines?

abc3: wdyt?

filipecabaco: fair... fair...

PR #353, opened 2024-06-09
docs: reformat documentation

hauleth:

Cleanup documentation formatting

Changes applied:

- Fixing some typos
- Apply 80 column limit on text lines
- Remove some tabs
- Remove some duplicated empty lines
- Capitalise SQL keywords


hauleth:
I used hard line limit on 80th column as it was easiest to automate. However it may be beneficial to use something like [semantic line breaks](https://sembr.org) if there is will in the team.

filipecabaco (approved):

PR #366, opened 2024-06-21
feat: validate username and db_name format

abc3:


abc3:
cc @hauleth

lib/supavisor/client_handler.ex
```diff
@@ -154,14 +154,16 @@ defmodule Supavisor.ClientHandler do
...
         {type, {user, tenant_or_alias, db_name}} = HH.parse_user_info(hello.payload)
 
-        not_allowed = ["\"", "\\"]
+        # Validate user and db_name according to PostgreSQL rules.
+        # The rules are: 1-63 characters, alphanumeric, underscore and $
+        rule = ~r/^[a-z_][a-z0-9_$]*$/
 
-        if String.contains?(user, not_allowed) or String.contains?(db_name, not_allowed) do
-          reason = "Invalid characters in user or db_name"
+        if !String.match?(user, rule) or !String.match?(db_name, rule) do
```
hauleth, line 161: I prefer using `not` over `!`. Also I think that `=~` would be cleaner there.

hauleth (approved):

lib/supavisor/client_handler.ex
abc3: > Also I think that =~ would be cleaner there.

good call!

abc3: > I prefer using not over !

is it just because of readability or is there some other benefit?

lib/supavisor/client_handler.ex
```diff
@@ -154,14 +154,16 @@ defmodule Supavisor.ClientHandler do
...
         {type, {user, tenant_or_alias, db_name}} = HH.parse_user_info(hello.payload)
 
-        not_allowed = ["\"", "\\"]
+        # Validate user and db_name according to PostgreSQL rules.
+        # The rules are: 1-63 characters, alphanumeric, underscore and $
+        rule = ~r/^[a-z_][a-z0-9_$]*$/
 
-        if String.contains?(user, not_allowed) or String.contains?(db_name, not_allowed) do
-          reason = "Invalid characters in user or db_name"
+        if !(user =~ rule && db_name =~ rule) do do
```
hauleth: I preferred it in older form

```suggestion
        if not user =~ rule or not db_name =~ rule do
```

hauleth: Alternatively we can swap bodies of `if` and use positive predicate there.

lib/supavisor/client_handler.ex
```diff
@@ -154,14 +154,16 @@ defmodule Supavisor.ClientHandler do
...
         {type, {user, tenant_or_alias, db_name}} = HH.parse_user_info(hello.payload)
 
-        not_allowed = ["\"", "\\"]
+        # Validate user and db_name according to PostgreSQL rules.
+        # The rules are: 1-63 characters, alphanumeric, underscore and $
+        rule = ~r/^[a-z_][a-z0-9_$]*$/
 
-        if String.contains?(user, not_allowed) or String.contains?(db_name, not_allowed) do
-          reason = "Invalid characters in user or db_name"
+        if !(user =~ rule && db_name =~ rule) do
```
hauleth, line 161: I preferred older form of:

```suggestion
        if not user =~ rule or not db_name =~ rule do
```

Though I would probably swap bodies of the `if` and use positive predicate there, so it would be:

```elixir
if user =~ rule and db_name =~ rule do
  # old `else` block content
else
  # old `if` block content
end
```

lib/supavisor/client_handler.ex
hauleth: 2 things:

- I find it more readable
- `!` works on any type, `not` works only on booleans

Technically `not` could be faster (though I doubt that it would be noticeable) and it also will warn you faster if there will be something off.

abc3: ! should be faster

abc3:
```
Name           ips        average  deviation         median         99th %
!          81.42 M       12.28 ns  ±1101.51%       12.08 ns       13.75 ns
not        78.34 M       12.77 ns  ±1097.19%       12.50 ns       14.58 ns

Comparison:
!          81.42 M
not        78.34 M - 1.04x slower +0.48 ns
```

```elixir
Benchee.run(%{
  "!" => fn ->
    !true
  end,
  "not" => fn ->
    not true
  end
})
```

lib/supavisor/client_handler.ex
abc3: 100%, just need to swap cases 😄

chasers (approved):

lib/supavisor/client_handler.ex
hauleth: So unnoticeable difference. Also I do not see how `!` can be faster, and there I think it could be optimised away by the compiler anyway.

PR #367, opened 2024-06-23
Use RPC helper everywhere

hauleth:
- **fix: add schema field to all migrations**
- **ft: use RPC helper function everywhere**
- **chore: bump version**


lib/supavisor/helpers.ex
```diff
@@ -329,12 +329,12 @@ defmodule Supavisor.Helpers do
...
     try do
       :erpc.call(node, module, function, args, timeout)
-      |> case do
-        {:EXIT, _} = badrpc -> {:error, {:badrpc, badrpc}}
-        result -> result
-      end
     catch
-      kind, reason -> {:error, {:badrpc, {kind, reason}}}
+      kind, reason ->
+        {:error, {:badrpc, {kind, reason}}}
```
abc3, line 334:
```suggestion
      kind, reason -> {:error, {:badrpc, {kind, reason}}}
```
```suggestion
      kind, reason ->
        {:error, {:badrpc, {kind, reason}}}
```

lib/supavisor_web/ws_proxy.ex
```diff
@@ -61,6 +61,6 @@ defmodule SupavisorWeb.WsProxy do
   @spec connect_local() :: {:ok, port()} | {:error, term()}
   defp connect_local() do
     proxy_port = Application.fetch_env!(:supavisor, :proxy_port_transaction)
-    :gen_tcp.connect('localhost', proxy_port, [:binary, packet: :raw, active: true])
+    :gen_tcp.connect(~c"localhost", proxy_port, [:binary, packet: :raw, active: true])
```
abc3, line 64: Is there any benefit to using sigils here?

abc3: ahh, it will be deprecated, right

hauleth: Yes, deprecation warnings.

lib/supavisor/helpers.ex
hauleth: I do not get this suggestions. You mean that it would be better to keep it in single line?

abc3: yeah

abc3: wdyt?

hauleth: Can be, other entries are also in single line, so it will be more consistent. Will push updated commit in a moment.

abc3 (approved):

PR #368, opened 2024-06-25
Move from `ct_slave` to `peer`

hauleth:

Replace deprecated `ct_slave` module with new `peer` module.

## Additional context

This also changes the sub-nodes to start only when needed instead of starting always for whole suite.

### DIFF
--- a/lib/supavisor/application.ex
+++ b/lib/supavisor/application.ex
@@ -34,19 +34,23 @@ defmodule Supavisor.Application do
     ]
 
     for {key, port, mode} <- proxy_ports do
-      :ranch.start_listener(
-        key,
-        :ranch_tcp,
-        %{
-          max_connections: String.to_integer(System.get_env("MAX_CONNECTIONS") || "25000"),
-          num_acceptors: String.to_integer(System.get_env("NUM_ACCEPTORS") || "100"),
-          socket_opts: [port: port, keepalive: true]
-        },
-        Supavisor.ClientHandler,
-        %{mode: mode}
-      )
-      |> then(&"Proxy started #{mode} on port #{port}, result: #{inspect(&1)}")
-      |> Logger.warning()
+      case :ranch.start_listener(
+             key,
+             :ranch_tcp,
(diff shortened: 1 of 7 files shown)


mix.exs
```diff
@@ -66,11 +66,12 @@ defmodule Supavisor.MixProject do
...
       # {:poolboy, "~> 1.5.2"},
       {:poolboy, git: "https://github.com/abc3/poolboy.git", tag: "v0.0.2"},
-      {:partisan, git: "https://github.com/lasp-lang/partisan.git", tag: "v5.0.0-rc.12"},
+      {:partisan, git: "https://github.com/lasp-lang/partisan.git", tag: "v5.0.0-rc.13"},
       {:syn, "~> 3.3"},
       {:pgo, "~> 0.13"},
-      {:rustler, "~> 0.29.1"}
+      {:rustler, "~> 0.29.1"},
       # TODO: add ranch deps
+      {:ranch, "~> 2.0", override: true}
```
abc3, line 74: are you sure that it's safe? I want to double check

mix.exs
```diff
@@ -66,11 +66,12 @@ defmodule Supavisor.MixProject do
       # pooller
       # {:poolboy, "~> 1.5.2"},
       {:poolboy, git: "https://github.com/abc3/poolboy.git", tag: "v0.0.2"},
-      {:partisan, git: "https://github.com/lasp-lang/partisan.git", tag: "v5.0.0-rc.12"},
+      {:partisan, git: "https://github.com/lasp-lang/partisan.git", tag: "v5.0.0-rc.13"},
```
abc3, line 69: let's not touch partisan yet, because it will be removed anyway

hauleth: I done it to support OTP 26 (though IIRC it still do not resolve an issue). I can roll it back.

mix.exs
hauleth: Also https://github.com/elixir-plug/plug_cowboy/pull/64

PR #369, opened 2024-06-27
ft: add struct for `ClientHandler` state

hauleth:

Adds struct for keeping `ClientHandler` state.

## Additional context

This approach allows newer Elixir versions to optimise code better, like for
example avoiding tests for whether passed argument is a map (it has to, as it
was already checked in the function head). It also allows compiler to catch at
compile time that you have accessed or you try to update incorrect field
(because of typo).


abc3:
Let's postpone this. It adds complexity with hot code upgrades, and I think the struct will add latency

PR #374, opened 2024-06-27
chore: reorganise GHA

hauleth:

- Use `erlef/setup-beam` action for setting up the tools instead of
  hand-crafted script
- Reorder steps in the CI to run lints faster to improve feedback loop


hauleth:
Ideally we could also split up CI action to run more things in parallel to improve feedback loop even more.

For example we could run:

- `mix format`
- `mix credo`
- Dialyzer checks
- `mix test`

in parallel to allow user to see where their code is failing faster, it also should result in successful runs happens faster, at the cost of more CI compute time spent.

PR #375, opened 2024-06-27
chore: split Elixir tasks

hauleth:

Split Elixir tasks in CI to improve build times.


.github/workflows/elixir.yml
```diff
@@ -0,0 +1,301 @@
...
+          key: ${{ runner.os }}-mix-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+          restore-keys: |
+            ${{ runner.os }}-mix-
+      - name: Run format check
+        run: mix format --check-formatted
+
+  credo:
+    name: Code style
+    runs-on: u22-arm-runner
+    needs: [compile-dev]
```
abc3, line 173: why do you want to check the dev build and not compile-test?

hauleth: Because Credo runs in `dev` environment there.

abc3: yeah, but probably we can run it for the test env and save a lot of GA time if we do not create a dev release

hauleth: We still create one for `mix dialyze`. But if you want we can converge these with `MIX_ENV=test` for everything.

hauleth: Also, this PR greatly improve caching, so that build times became less of a problem.

mix.lock
```diff
@@ -13,20 +13,20 @@
...
   "hpax": {:hex, :hpax, "0.1.2", "09a75600d9d8bbd064cdd741f21fc06fc1f4cf3d0fcc335e5aa19be1a7235c84", [:mix], [], "hexpm", "2c87843d5a23f5f16748ebe77969880e29809580efdaccd615cd3bed628a8c13"},
-  "inet_cidr": {:hex, :inet_cidr, "1.0.4", "a05744ab7c221ca8e395c926c3919a821eb512e8f36547c062f62c4ca0cf3d6e", [:mix], [], "hexpm", "64a2d30189704ae41ca7dbdd587f5291db5d1dda1414e0774c29ffc81088c1bc"},
+  "inet_cidr": {:hex, :inet_cidr, "1.0.8", "d26bb7bdbdf21ae401ead2092bf2bb4bf57fe44a62f5eaa5025280720ace8a40", [:mix], [], "hexpm", "d5b26da66603bb56c933c65214c72152f0de9a6ea53618b56d63302a68f6a90e"},
```
abc3, line 29: let's upgrade the apps in the supavisor  v2

hauleth: I did it for testing purposes. I can revert it if you want.

.github/workflows/elixir.yml
abc3: > We still create one for mix dialyze. But if you want we can converge these with MIX_ENV=test for everything.
Let's try to do that. Basically, there is no difference between the dev and test environments

mix.lock
abc3: Yes, please

abc3: don't want to block the current main branch for hotfixes

hauleth: Sure. Will revert in a second.

.github/workflows/elixir.yml
hauleth: Now everything is in `test` environment.

.github/workflows/prod.yml
```diff
@@ -14,28 +14,27 @@ jobs:
...
-        run: |
-            . ~/.asdf/asdf.sh
-            echo 'elixir v1.14.5-otp-25' >> ~/.tool-versions
-            echo 'erlang 25.3.2.7' >> ~/.tool-versions
-            elixir -v
+        uses: erlef/setup-beam@v1
+        with:
+          otp-version: '25.3.2.7'
+          elixir-version: '1.17'
       - name: Set up Rust
```
abc3, line 32: could you revert this as well? we need to build the tarball with the same Erlang and Elixir versions as set up on our servers

.github/workflows/stage.yml
```diff
@@ -14,28 +14,27 @@ jobs:
...
       - name: Setup Elixir
-        run: |
-            . ~/.asdf/asdf.sh
-            echo 'elixir v1.14.5-otp-25' >> ~/.tool-versions
-            echo 'erlang 25.3.2.7' >> ~/.tool-versions
-            elixir -v
+        uses: erlef/setup-beam@v1
+        with:
+          otp-version: '25.3.2.7'
+          elixir-version: '1.17'
```
abc3, line 31: and this

.github/workflows/prod.yml
hauleth: Surę, I will update it to use the previous versions.

.github/workflows/prod.yml
```diff
@@ -14,28 +14,27 @@ jobs:
...
       - name: Setup Elixir
-        run: |
-            . ~/.asdf/asdf.sh
-            echo 'elixir v1.14.5-otp-25' >> ~/.tool-versions
-            echo 'erlang 25.3.2.7' >> ~/.tool-versions
-            elixir -v
+        uses: erlef/setup-beam@v1
+        with:
+          otp-version: '25.3.2.7'
+          elixir-version: '1.14.5'
```
abc3, line 31: Please return it to its original state. [That doesn't work with the customer runner](https://github.com/supabase/supavisor/pull/226/files) and such changes require a sensitive and careful check because they can break the current deployment process.

hauleth:
Some problems with Dialyzer when ran in `test` environment - https://github.com/supabase/supavisor/actions/runs/9711545084/job/26804706396?pr=375#step:10:31

abc3:
we can ignore it via .dialyzer_ignore.exs

abc3 (approved):
🔥 🔥 🔥

PR #376, opened 2024-06-28
chore: run CI tests when targeting `v2` branch

hauleth:


abc3 (approved):

hauleth:
The additional compilation step was too flaky to be useable, so I have removed it and instead cached just Mix deps and Cargo output. It provides enough CI-time optimisation.

PR #392, opened 2024-07-08
ft: remove all HTML-related functionalities

hauleth:

Supavisor is not using any HTML views, it do not offer any HTML-based UI, etc. so these functionalities aren't needed for anything.


hauleth:
I have tested it and Swagger works. I also have spotted some issues that I am fixing in #393

PR #401, opened 2024-07-17
feat: implement proxy handler

abc3:
The current PR introduces a new handler that will replace session mode. It authenticates incoming connections via auth queries, establishes a TCP connection to the tenant's database, and then simply forwards data between the client and the database, functioning primarily as a proxy

- [x] cancel query
- [x] terminate auth_query connection
- [x] active once
- [x] application_name with "via supavisor"
- [x] handle client disconnect
- [x] handle DB disconnect
- [x] auth_query conn is terminated if the client is killed
- [x] telemetry
- [x] delete metrics  
- [x] tests

### DIFF
--- a/lib/supavisor/handlers/proxy/client.ex
+++ b/lib/supavisor/handlers/proxy/client.ex
@@ -1,0 +1,662 @@
+defmodule Supavisor.Handlers.Proxy.Client do
+  @moduledoc false
+
+  require Logger
+
+  alias Supavisor, as: S
+  alias Supavisor.ProxyDb, as: Db
+  alias Supavisor.Helpers, as: H
+  alias Supavisor.HandlerHelpers, as: HH
+
+  alias Supavisor.{
+    Tenants,
+    ProxyHandlerDb,
+    Monitoring.Telem,
+    Protocol.Client,
+    Protocol.Server
+  }
+
+  alias Supavisor.Handlers.Proxy.Db, as: ProxyDb
+
+  @cancel_query_msg <<16::32, 1234::16, 5678::16>>
+  @sock_closed [:tcp_closed, :ssl_closed]
+  @proto [:tcp, :ssl]
+
+  def handle_event(:info, {_proto, _, <<"GET", _::binary>>}, :exchange, data) do
+    Logger.debug("ProxyClient: Client is trying to request HTTP")
+    HH.sock_send(data.sock, "HTTP/1.1 204 OK\r\nx-app-version: #{data.version}\r\n\r\n")
+    {:stop, {:shutdown, :http_request}}
+  end
+
+  # cancel request
+  def handle_event(:info, {_, _, <<@cancel_query_msg, pid::32, key::32>>}, _, _) do
+    Logger.debug("ProxyClient: Got cancel query for #{inspect({pid, key})}")
+    :ok = HH.send_cancel_query(pid, key, {:client, :cancel_query})
+    {:stop, {:shutdown, :cancel_query}}
+  end
+
+  def handle_event(:info, {:client, :cancel_query}, _, %{
+        auth: auth,
+        backend_key_data: b
+      }) do
+    :ok = HH.cancel_query(~c"#{auth.host}", auth.port, auth.ip_ver, b.pid, b.key)
+    :keep_state_and_data
+  end
+
+  # # ssl request from client
+  def handle_event(:info, {:tcp, _, <<_::64>>}, :exchange, %{sock: sock} = data) do
+    Logger.debug("ProxyClient: Client is trying to connect with SSL")
+
+    downstream_cert = H.downstream_cert()
+    downstream_key = H.downstream_key()
+
+    # SSL negotiation, S/N/Error
+    if !!downstream_cert and !!downstream_key do
+      :ok = HH.setopts(sock, active: false)
+      :ok = HH.sock_send(sock, "S")
+
+      opts = [
+        certfile: downstream_cert,
+        keyfile: downstream_key
+      ]
+
+      case :ssl.handshake(elem(sock, 1), opts) do
+        {:ok, ssl_sock} ->
+          socket = {:ssl, ssl_sock}
+          :ok = HH.setopts(socket, active: true)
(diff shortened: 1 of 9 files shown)


lib/supavisor/handlers/proxy/handler.ex
```diff
@@ -0,0 +1,155 @@
...
+      connection_start: System.monotonic_time(),
+      log_level: nil,
+      version: Application.spec(:supavisor, :vsn),
+      auth: nil,
+      nonce: nil,
+      server_proof: nil,
+      parameter_status: %{},
+      app_name: nil,
+      peer_ip: Helpers.peer_ip(sock),
+      auth: %{},
```
hauleth, line 74: `auth` key is mentioned twice.

lib/supavisor/handlers/proxy/handler.ex
```diff
@@ -0,0 +1,155 @@
...
+
+  ## Internal functions
+
+  @spec clean_up(map()) :: any()
+  defp clean_up(data) do
+    HH.sock_close(data.sock)
+    HH.sock_close(data.db_sock)
+
+    case Registry.lookup(Supavisor.Registry.TenantClients, data.id) do
+      clients when clients in [[{self(), []}], []] -> PromEx.remove_metrics(data.id)
```
hauleth, line 151: I have some issues with readability of this line. I do not know how to make it more readable, maybe something like:

```elixir
if List.first(lookup_result, {self(), []}) == {self(), []} do
  # …
```

Would be more readable?

lib/supavisor/handlers/proxy/handler.ex
```diff
@@ -0,0 +1,155 @@
...
+
+  alias Supavisor.HandlerHelpers, as: HH
+
+  @sock_closed [:tcp_closed, :ssl_closed]
+  @proto [:tcp, :ssl]
+
+  @impl true
+  def start_link(ref, transport, opts) do
+    pid = :proc_lib.spawn_link(__MODULE__, :init, [ref, transport, opts])
+    {:ok, pid}
```
hauleth, line 25: I would use `:proc_lib.start_link/3` instead.

lib/supavisor/handlers/proxy/client.ex
```diff
@@ -0,0 +1,662 @@
...
+
+            Telem.client_join(:fail, id)
+            {:stop, {:shutdown, :address_not_allowed}}
+
+          true ->
+            new_data = update_user_data(data, info, user, id, db_name, mode)
+
+            key = {:secrets, tenant_or_alias, user}
+
+            case auth_secrets(info, user, key, :timer.hours(24)) do
```
hauleth, line 175: Deeply nested conditionals make it hard to follow the flow of the code.

lib/supavisor/handlers/proxy/client.ex
```diff
@@ -0,0 +1,662 @@
...
+      {:error, reason} ->
+        Logger.error(
+          "ProxyClient: User not found: #{inspect(reason)} #{inspect({type, user, tenant_or_alias})}"
+        )
+
+        :ok = HH.send_error(sock, "XX000", "Tenant or user not found")
+        Telem.client_join(:fail, data.id)
+        {:stop, {:shutdown, :user_not_found}}
+    end
+  end
```
hauleth, line 202: This function is quite big.

lib/supavisor/handlers/proxy/client.ex
```diff
@@ -0,0 +1,662 @@
...
+            do: Server.error_message("XX000", reason),
+            else: Server.exchange_message(:final, "e=#{reason}")
+
+        key = {:secrets_check, data.tenant, data.user}
+
+        if method != :password and reason == "Wrong password" and
+             Cachex.get(Supavisor.Cache, key) == {:ok, nil} do
+          case auth_secrets(info, data.user, key, 15_000) do
+            {:ok, {method2, secrets2}} = value ->
+              if method != method2 or Map.delete(secrets.(), :client_key) != secrets2.() do
```
hauleth, line 229: Deeply nested conditionals again.

lib/supavisor/handlers/proxy/client.ex
```diff
@@ -0,0 +1,662 @@
...
+      ip_ver
+    ]
+
+    case :gen_tcp.connect(~c"#{auth.host}", auth.port, sock_opts) do
+      {:ok, sock} ->
+        Logger.debug("ProxyClient: auth #{inspect(auth, pretty: true)}")
+
+        case ProxyDb.try_ssl_handshake({:gen_tcp, sock}, auth) do
+          {:ok, sock} ->
+            case ProxyDb.send_startup(sock, auth) do
```
hauleth, line 294: Deeply nested conditional.

hauleth (approved):
Some compilation warnings to cleanup, and some readability issues that I have found.

hauleth (changes requested):
I have clicked incorrectly previous time.

lib/supavisor/handlers/proxy/handler.ex
abc3: Any reason for that? In the cowboy codebase, they mostly use spawn_link

lib/supavisor/handlers/proxy/handler.ex
abc3: for me, this variant is harder to read :D

abc3:
nested or long functions can be simplified before merging v2 into main, once the handler for transaction mode is ready

lib/supavisor/handlers/proxy/handler.ex
hauleth: Somehow `start_link/3` is cleaner for me, as I can return error from spawned process (and also inform the parent when the handler is ready).

lib/supavisor/handlers/proxy/handler.ex
hauleth: Alternatively we could expand the guard to full form `when clients == [{self(), []}] or clients == []`

abc3: yeah, that is good

PR #402, opened 2024-07-18
feat: add new transaction handler

abc3:


lib/supavisor/handlers/proxy/handler.ex
```diff
@@ -146,9 +145,13 @@ defmodule Supavisor.Handlers.Proxy.Handler do
     HH.sock_close(data.sock)
     HH.sock_close(data.db_sock)
 
-    case Registry.lookup(Supavisor.Registry.TenantClients, data.id) do
-      clients when clients == [{self(), []}] or clients == [] -> PromEx.remove_metrics(data.id)
-      _ -> :ok
+    if data.id != nil do
+      with clients <- Registry.lookup(Supavisor.Registry.TenantClients, data.id),
```
hauleth, line 149: Why use `with` for this clause? It cannot fail.

lib/supavisor.ex
```diff
@@ -349,4 +349,32 @@ defmodule Supavisor do
...
+  end
+
+  @spec start_local_server(map()) :: {:ok, map()} | {:error, any()}
+  def start_local_server(args) do
+    acceptors = round(args.max_clients / 100)
+
+    opts =
+      %{
+        max_connections: args.max_clients,
+        num_acceptors: if(acceptors < 10, do: 10, else: acceptors),
```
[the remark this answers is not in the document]
abc3: ohh, cool, didn't know  that

lib/supavisor/handlers/proxy/client.ex
```diff
@@ -4,20 +4,17 @@ defmodule Supavisor.Handlers.Proxy.Client do
   require Logger
 
   alias Supavisor, as: S
-  alias Supavisor.ProxyDb, as: Db
   alias Supavisor.Helpers, as: H
   alias Supavisor.HandlerHelpers, as: HH
 
-  alias Supavisor.{
+  alias S.{
```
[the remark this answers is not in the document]
abc3: yeah, let's make it more readable

lib/supavisor/handlers/proxy/handler.ex
abc3: :facepalm:

lib/supavisor.ex
```diff
@@ -349,4 +349,36 @@ defmodule Supavisor do
...
+      _ -> {:error, :not_found}
+    end
+  end
+
+  @spec start_local_server(map()) :: {:ok, map()} | {:error, any()}
+  def start_local_server(%{max_clients: max_clients} = args) do
+    # max_clients=-1 is used for testing the maximum allowed clients in ProxyTest
+    {acceptors, max_clients} =
+      if max_clients > 0,
+        do: {round(max_clients / 100), max_clients},
```
hauleth, line 366: What if `max_clients` is smaller than `50`? We will end with no `acceptors`?

hauleth (approved):

lib/supavisor.ex
abc3: right, must be `ceil`

lib/supavisor/db_handler.ex
```diff
@@ -31,6 +31,12 @@ defmodule Supavisor.DbHandler do
   @spec cast(pid(), pid(), binary()) :: :ok | {:error, any()} | {:buffering, non_neg_integer()}
   def cast(pid, caller, msg), do: :gen_statem.cast(pid, {:db_cast, caller, msg})
 
+  @spec set_idle(pid()) :: :ok
+  def set_idle(pid), do: :gen_statem.cast(pid, :set_idle)
+
+  @spec change_skt_owner(pid(), pid()) :: {:ok, S.sock()}
+  def change_skt_owner(pid, caller), do: :gen_statem.call(pid, {:tcp_owner, caller}, 15_000)
```
hauleth, line 38:
```suggestion
  @spec change_socket_owner(pid(), pid()) :: {:ok, S.sock()}
  def change_socket_owner(pid, caller), do: :gen_statem.call(pid, {:tcp_owner, caller}, 15_000)
```

Do not shorten name as it make it less readable.

hauleth: If you want to shorten it, then `sock` would be IMHO better than `skt`, and it is only 1 character longer.

PR #408, opened 2024-07-31
Supavisor V2

abc3:
In this PR, Supavisor has been adapted for Erlang 27 and Elixir 1.17. The metrics backend has been changed from [PromEx](https://github.com/akoutmos/prom_ex) to [Peep](https://github.com/rkallos/peep), and the most significant change is the introduction of [ProxyHandler](https://github.com/supabase/supavisor/tree/v2/lib/supavisor/handlers/proxy) for processing transaction mode.

The ProxyHandler first authenticates the incoming connection through auth_query. If there is no connection pool, it creates one and also launches a [Ranch on a free port](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor.ex#L362-L383), saving the port and host in [the corresponding Syn record](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/tenant_supervisor.ex#L11).

If the pool is on the same node as the client connection, we locally [change the owner of the TCP socket](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/db_handler.ex#L407-L416) for each transaction to the direct connection with the database. After [the transaction is completed](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/handlers/proxy/db.ex#L75-L90), the client process returns ownership of the TCP socket to the DbHandler.

[If the pool is on a different node](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/handlers/proxy/client.ex#L313), the client process connects to the Ranch of the corresponding pool on the other node, [switches to session mode](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/handlers/proxy/client.ex#L335-L349), and forwards everything (works like proxy). The local Ranch operates the same way as when the client connection is on the same node as the connection pool.

With this handler, erldist is used only for synchronizing Syn. Forwarding query data between nodes is done via TCP sockets (Ranch)


abc3:
Hey, @supabase/dashbit, we have prepared improvements for Supavisor and would love to hear your feedback. There are a lot of changes, so feel free to review whatever you like. We are particularly interested in your opinion on the [ProxyHandler](https://github.com/supabase/supavisor/tree/v2/lib/supavisor/handlers/proxy). The corresponding implementation is here:

- [lib/supavisor/handlers/proxy/handler.ex](https://github.com/supabase/supavisor/blob/v2/lib/supavisor/handlers/proxy/handler.ex)
- [lib/supavisor/handlers/proxy/client.ex](https://github.com/supabase/supavisor/blob/v2/lib/supavisor/handlers/proxy/client.ex)
- [lib/supavisor/handlers/proxy/db.ex](https://github.com/supabase/supavisor/blob/v2/lib/supavisor/handlers/proxy/db.ex)

josevalim:
Exciting work! Do you have any numbers or results into the performance of this new version?

> After [the transaction is completed](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/handlers/proxy/db.ex#L75-L90), the client process returns ownership of the TCP socket to the DbHandler.

I haven't reviewed the code yet, so I will try to verify this as well, but remember that, if the client crashes for some reason, you most likely can't assert for certain the state of the socket: for example, was something only partially written to it? So you may need to drop the connection altogether or do some sort of cleanup (if possible), like make sure to abort transactions and other things. I am not sure how much state can be carried around.

josevalim:
Two additional questions for now:

1. I also remember we discussed improvements to poolboy or your fork to poolboy. I don't remember what they were exactly though. I assume those may already be in place?

2. Is it correct that this pull request still preserves the old code for talking to the database? Is the plan to remove it in the future or are there issues in unifying everything through the proxy handler?

josevalim:
Two more:

> The ProxyHandler first authenticates the incoming connection through auth_query. If there is no connection pool, it creates one and also launches a [Ranch on a free port](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor.ex#L362-L383), saving the port and host in [the corresponding Syn record](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/tenant_supervisor.ex#L11).

3. Is the maximum number of ports (65k) a potential concern here? If so, could we use a single Ranch port but pass some other information (such as a preamble when connecting via gen_tcp) to choose with pool to use?

> [If the pool is on a different node](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/handlers/proxy/client.ex#L313), the client process connects to the Ranch of the corresponding pool on the other node, [switches to session mode](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/handlers/proxy/client.ex#L335-L349), and forwards everything (works like proxy).

4. How much do you handoff to the other node? From looking at the code, if it is on a different node, you are doing both auth and ssl handshakes. Does it mean you are doing them twice? On in the current node and then once you have to proxy? I wonder if it is worth doing a "special handshake" between those nodes, so when you join this special Ranch port (which should not be exposed), you can assume auth and ssl have been handled in the first node and you don't have to redo it. Maybe that's already the case, but I wanted to double check.

abc3:
> Exciting work!

:blush::blush::blush:

> Do you have any numbers or results into the performance of this new version?

We are still measuring performance for different scenarios. When the pooler is on the same machine as the database, Supavisor outperforms or performs the same as PgBouncer. However, it is slightly slower when a Supavisor cluster of two nodes is in the same AWS zone as the database.


> I haven't reviewed the code yet, so I will try to verify this as well, but remember that, if the client crashes for some reason, you most likely can't assert for certain the state of the socket: for example, was something only partially written to it? So you may need to drop the connection altogether or do some sort of cleanup (if possible), like make sure to abort transactions and other things. I am not sure how much state can be carried around.

We keep a linked DbHandler which originally created the socket, and if the client goes down, DbHandler will be respawned with a new direct connection.

> I also remember we discussed improvements to poolboy or your fork to poolboy. I don't remember what they were exactly though. I assume those may already be in place?

I spent some time finding bottlenecks, and the biggest one was metrics collection (shout out to @hauleth who fixed that) and the synchronous calling between ClientHandler and DbHandler to avoid overwhelming the mailboxes. So far, we haven't added anything new to Poolboy, except our previous changes for idle timeout.

> Is it correct that this pull request still preserves the old code for talking to the database? Is the plan to remove it in the future or are there issues in unifying everything through the proxy handler?

DbHandler is still responsible for establishing direct connections in transaction mode, reconnecting in idle state, and it has been extended with handler_call to change ownership of the TCP socket. However, the connection logic for talking with the database during transactions is [located in ProxyDb](https://github.com/supabase/supavisor/blob/52191dae8ce6a7fcff67ca87dc70ebe4b5e6f0c0/lib/supavisor/handlers/proxy/db.ex#L70-L96). In the future, we will remove the old module.

> Is the maximum number of ports (65k) a potential concern here? If so, could we use a single Ranch port but pass some other information (such as a preamble when connecting via gen_tcp) to choose with pool to use?

Not at the moment, but it could potentially become a problem. The update is already quite complex, so I don't want to add further complications. We can always add this feature later


> How much do you handoff to the other node? From looking at the code, if it is on a different node, you are doing both auth and ssl handshakes. Does it mean you are doing them twice? On in the current node and then once you have to proxy? I wonder if it is worth doing a "special handshake" between those nodes, so when you join this special Ranch port (which should not be exposed), you can assume auth and ssl have been handled in the first node and you don't have to redo it. Maybe that's already the case, but I wanted to double check.

Hmm, it should not use SSL for proxying to another node. The main reason for authentication was to provide an additional security guard. It's not very resource-intensive but can prevent accidental connections if the client has some issues with authentication.

josevalim:
Fantastic. I have already reviewed the code and I will do another pass later. My only suggestion so far would be to have a custom tcp handshake for the proxy, so you have only a single special port instead of a pool of them and there is no need for additional auth. For security, you can put a 20-40 bytes random token on Syn instead of the Ranch port, and the new handshake uses a key plus this token for connecting. But as you said, I don't think it is a necessary change at this moment.

> We are still measuring performance for different scenarios. When the pooler is on the same machine as the database, Supavisor outperforms or performs the same as PgBouncer. However, it is slightly slower when a Supavisor cluster of two nodes is in the same AWS zone as the database.

This is great! As far as I know, no other "bouncer" offers clustering anyway, and slightly slower would certainly be expected.

abc3:
> My only suggestion so far would be to have a custom tcp handshake for the proxy, so you have only a single special port instead of a pool of them and there is no need for additional auth. For security, you can put a 20-40 bytes random token on Syn instead of the Ranch port, and the new handshake uses a key plus this token for connecting

a good call, thanks!

lib/supavisor/handlers/proxy/client.ex
```diff
@@ -0,0 +1,784 @@
...
+  def app_name(name) when is_binary(name) do
+    suffix = " via Supavisor"
+    # https://www.postgresql.org/docs/current/runtime-config-logging.html#GUC-APPLICATION-NAME
+    max_len = 64
+    suffix_len = 14
+
+    if String.length(name) <= max_len - suffix_len do
+      name <> suffix
+    else
+      truncated_name = String.slice(name, 0, max_len - suffix_len - 3)
```
josevalim, line 745: I assume the limit is in bytes, not in UTF8 codepoints. While perhaps no one will have UTF8 codepoints in their application name, better to be safe:

```suggestion
    if byte_size(name) <= max_len - suffix_len do
      name <> suffix
    else
      truncated_name = binary_slice(name, 0, max_len - suffix_len - 3)
```

lib/supavisor/handlers/proxy/client.ex
```diff
@@ -0,0 +1,784 @@
...
+      "ProxyClient: Connected to db #{tenant.db_host} #{tenant.db_port} #{tenant.db_database} #{user.db_user}"
+    )
+
+    resp =
+      with {:ok, secret} <- Helpers.get_user_secret(conn, tenant.auth_query, db_user) do
+        t = if secret.digest == :md5, do: :auth_query_md5, else: :auth_query
+        {:ok, {t, fn -> Map.put(secret, :alias, user.db_user_alias) end}}
+      end
+
+    GenServer.stop(conn, :normal, 5_000)
```
josevalim, line 670: Could there be a bug here? Once you stop the process, it will deliver a `{:EXIT, pid, :normal}` message to this process, since you are trapping exits. I don't see such messages being discarded but I may have missed where it happens, so I thought about dropping a comment.

Similarly, if whatever you check out from poolboy crashes, you may also get an EXIT message, although this case can only happen if the pool terminates, so I assume something else is already going wrong.

josevalim (commented):
I did another review of the proxy modules and I have dropped only two additional comments. :)

lib/supavisor/handlers/proxy/client.ex
abc3: This code occurs [inside Cachex's fetch](https://github.com/supabase/supavisor/blob/6e54020d307caee55f16d4ef833ed03fa9aa3cc9/lib/supavisor/handlers/proxy/client.ex#L613-L626), so it's [linked to an internally spawned process](https://github.com/supabase/supavisor/blob/6e54020d307caee55f16d4ef833ed03fa9aa3cc9/lib/supavisor/handlers/proxy/client.ex#L657-L658)

hauleth (approved):

hauleth:
Moved `v2` to `main`.

PR #411, opened 2024-08-02
feat: disable metrics via environment variable

abc3:


lib/supavisor/monitoring/telem.ex
```diff
@@ -3,69 +3,102 @@ defmodule Supavisor.Monitoring.Telem do
 
   require Logger
 
-  alias Supavisor, as: S
+  @metrics_disabled Application.compile_env(:supavisor, :metrics_disabled, false)
 
-  @spec network_usage(:client | :db, S.sock(), S.id(), map()) :: {:ok | :error, map()}
+  defmacro if_metrics_enabled(do: block) do
```
hauleth, line 8: I think that instead of creating such macro, that applies code only if the metrics are enabled, I would create macro:

```elixir
defmacro telemetry_execute(event_name, measurements, metadata) do
  if not @metrics_disabled do
    quote do
      :telemetry.execute(unquote(event_name), unquote(measurements), unquote(metadata))
    end
  end
end
```

That would IMHO be easier to read.

lib/supavisor/monitoring/telem.ex
```diff
@@ -3,69 +3,102 @@ defmodule Supavisor.Monitoring.Telem do
...
+        {:ok, %{recv_oct: 0, send_oct: 0}}
+      end
+    else
+      block
+    end
+  end
+
+  @spec network_usage(:client | :db, Supavisor.sock(), Supavisor.id(), map()) ::
+          {:ok | :error, map()}
   def network_usage(type, {mod, socket}, id, _stats) do
```
hauleth, line 30: I think that I would put whole function inside of `if` and if disabled, then I would create NOP function.

lib/supavisor/monitoring/telem.ex
abc3: yeah, agree

abc3:
and yeah we need to agree to avoid using `unless` in favor of `if not`

PR #413, opened 2024-08-02
style: clean Credo warnings

hauleth:


lib/supavisor/handlers/proxy/client.ex
```diff
@@ -378,7 +379,7 @@ defmodule Supavisor.Handlers.Proxy.Client do
             case Db.send_startup(sock, auth, tenant) do
               :ok ->
                 HandlerHelpers.active_once(sock)
-                auth = Map.put(auth, :ip_ver, ip_ver)
+                auth = Map.put(auth, :ip_version, ip_ver)
```
abc3, line 382: There are :inet or :inet6 in ip_ver, and :ip4 or :ip6 in ip_version

hauleth: That change wasn't meant to be there, so I will clean that up and update.

abc3: I will put data in defstruct, so it will be easy to refactor

abc3 (approved):

PR #416, opened 2024-08-05
fix: active true in db_handler

abc3:


lib/supavisor/handlers/proxy/db.ex
```diff
@@ -110,6 +123,19 @@ defmodule Supavisor.Handlers.Proxy.Db do
 
   ## Internal functions
 
+  @spec maybe_active_once(map) :: non_neg_integer()
+  def maybe_active_once(%{db_send: 0} = data) do
+    HandlerHelpers.setopts(data.db_sock, active: true)
```
hauleth, line 128: Why use `active: true`? I think it could create issues related to the messages overload with no back pressure on the sender.

abc3: this pr is for debugging

PR #419, opened 2024-08-12
feat: directly send to socket

abc3:


Makefile
```diff
@@ -105,14 +106,15 @@ dev_start_rel:
 	DB_POOL_SIZE="5" \
 	_build/prod/rel/supavisor/bin/supavisor start_iex
 
+# rm -rf _build/prod && \
+
```
hauleth, line 110:
```suggestion
```

lib/supavisor/client_handler.ex
```diff
@@ -382,12 +395,48 @@ defmodule Supavisor.ClientHandler do
         Telem.client_join(:fail, data.id)
         {:stop, {:shutdown, :max_clients_reached}}
 
+      :proxy ->
+        {:ok, %{port: port, host: host}} = Supavisor.get_pool_ranch(data.id)
+
+        auth =
+          Map.merge(data.auth, %{
+            port: port,
+            host: ~c"#{host}",
```
hauleth, line 404:
```suggestion
            host: to_charlist(host),
```

lib/supavisor/client_handler.ex
```diff
@@ -382,12 +395,48 @@ defmodule Supavisor.ClientHandler do
...
+        {:keep_state, %{data | auth: auth}, {:next_event, :internal, :connect_db}}
+
       error ->
         Logger.error("ClientHandler: Subscribe error: #{inspect(error)}")
         {:keep_state_and_data, {:timeout, 1000, :subscribe}}
     end
   end
 
+  def handle_event(:internal, :connect_db, _, data) do
+    Logger.debug("Try to connect to DB")
```
hauleth, line 420:
```suggestion
    Logger.debug("ClientHandler: Trying to connect to DB")
```

Nitpick, but it makes more sense as it can be read as "I am trying…".

lib/supavisor/client_handler.ex
```diff
@@ -783,32 +797,49 @@ defmodule Supavisor.ClientHandler do
...
   defp db_checkout(_, _, data) do
-    {time, db_pid} = :timer.tc(:poolboy, :checkout, [data.pool, true, data.timeout])
+    start = System.monotonic_time(:microsecond)
+    db_pid = :poolboy.checkout(data.pool, true, data.timeout)
     Process.link(db_pid)
+    db_sock = DbHandler.checkout(db_pid, data.sock, self())
     same_box = if node(db_pid) == node(), do: :local, else: :remote
-    Telem.pool_checkout_time(time, data.id, same_box)
-    {data.pool, db_pid}
+    Telem.pool_checkout_time(System.monotonic_time(:microsecond) - start, data.id, same_box)
```
hauleth, line 805: I recall that somewhere in the code I have created macro that runs something similar to this example. Maybe it would be useful to reuse such macro to make intent clearer there.

lib/supavisor/db_handler.ex
```diff
@@ -86,13 +82,19 @@ defmodule Supavisor.DbHandler do
...
+        packet: :raw,
+        # recbuf: 8192,
+        # sndbuf: 8192,
+        # backlog: 2048,
+        # send_timeout: 120,
+        # keepalive: true,
+        # nopush: true,
+        nodelay: true,
+        active: true
+      ] ++ [auth.ip_version]
```
hauleth, line 97: Why use `++/2` there? We can put `auth.ip_version` on top, and it will work without such workaround.

lib/supavisor/db_handler.ex
```diff
@@ -135,166 +139,64 @@ defmodule Supavisor.DbHandler do
...
-
-          payload = ["md5", Helpers.md5([digest, salt]), 0]
-          bin = [?p, <<IO.iodata_length(payload) + 4::signed-32>>, payload]
-          :ok = sock_send(data.sock, bin)
-          {ps, :authentication_md5}
-
-        %{tag: :error_response, payload: error}, _ ->
-          {:error_response, error}
+    Logger.debug("ProxyDb: dec_pkt, #{inspect(dec_pkt, pretty: true)}")
+    # HandlerHelpers.active_once(data.sock)
```
hauleth, line 143:
```suggestion
```

lib/supavisor/db_handler.ex
```diff
@@ -135,166 +139,64 @@ defmodule Supavisor.DbHandler do
...
         {:stop, :auth_error, data}
     end
   end
 
-  def handle_event(:internal, :check_buffer, :idle, %{reply: {from, pid}} = data) do
-    :ok = Helpers.controlling_process(data.sock, pid)
-    reply = {:reply, from, {:ok, data.sock}}
-    {:next_state, :busy, %{data | reply: nil}, reply}
+  def handle_event(:internal, :check_buffer, :idle, %{reply: from} = data) when from != nil do
+    Logger.debug("DbHandler: Check buffer")
```
hauleth, line 198: It should be `ProxyDb` or `DbHandler`? Because it is mixed in the different log entries.

lib/supavisor/db_handler.ex
```diff
@@ -352,44 +261,38 @@ defmodule Supavisor.DbHandler do
...
-      case ready do
-        {:ready_for_query, :idle} -> {:client_cast, :ready_for_query}
-        {:ready_for_query, _} -> {:client_cast, :continue}
-        _ when sent < @async_send_limit -> {:client_cast, :continue}
-        _ -> {:client_call, :continue}
-      end
+    Logger.debug("DbHandler: Got write replica message  #{inspect(bin)}")
 
-    :ok = apply(ClientHandler, send_via, [data.caller, bin, progress])
+    # HandlerHelpers.setopts(data.sock, active: :once)
```
hauleth, line 278:
```suggestion
```

lib/supavisor/db_handler.ex
```diff
@@ -477,7 +338,7 @@ defmodule Supavisor.DbHandler do
     end
 
     if state == :busy or data.mode == :session do
-      sock_send(data.sock, <<?X, 4::32>>)
+      # sock_send(data.sock, <<?X, 4::32>>)
```
hauleth, line 341:
```suggestion
```

lib/supavisor/db_handler.ex
abc3: DbHandler, ProxyDb is a legacy from the another handler

lib/supavisor/client_handler.ex
abc3: what is the difference?

lib/supavisor/client_handler.ex
hauleth: NVM, I have written it, but it never landed.

PR #422, opened 2024-08-22
feat: ability to determine the pool node by aws zone

abc3:


config/runtime.exs
```diff
@@ -159,7 +159,8 @@ if config_env() != :test do
     reconnect_on_db_close: System.get_env("RECONNECT_ON_DB_CLOSE") == "true",
     api_blocklist: System.get_env("API_TOKEN_BLOCKLIST", "") |> String.split(","),
     metrics_blocklist: System.get_env("METRICS_TOKEN_BLOCKLIST", "") |> String.split(","),
-    node_host: System.get_env("NODE_IP", "127.0.0.1")
+    node_host: System.get_env("NODE_IP", "127.0.0.1"),
+    aws_zone: System.get_env("AWS_ZONE")
```
hauleth, line 163: Couldn't we use [AWS Instance Metadata Service](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMI-launch-index-examples.html) to fetch that instead of relying on configuration?

hauleth (approved):

config/runtime.exs
abc3: During the EC2 setup, our Pulumi script will retrieve this information and place it in an environment variable for the systemd supavisor service. This prevents Supavisor from having to wait for an AWS metadata response during startup or reboot.

PR #451, opened 2024-09-27
fix: handling of active_count

abc3:


chasers:
How do we have a test for this?

abc3:
for example, add a test for session and transaction modes where the transaction contains 100+ queries

lib/supavisor/db_handler.ex
```diff
@@ -297,8 +297,10 @@ defmodule Supavisor.DbHandler do
     if String.ends_with?(bin, Server.ready_for_query()) do
       {_, stats} = Telem.network_usage(:db, data.sock, data.id, data.stats)
 
+      # in transaction mode, we need to notify the client when the transaction is finished,
+      # after which it will unlink the direct db connection process from itself.
       data =
-        if data.mode in [:transaction, :session] do
+        if data.mode == :transaction do
```
hauleth, line 303: Wasn't that change added recently? It is reverting it?

hauleth (approved):

lib/supavisor/db_handler.ex
abc3: Yes, but we don't need it. The ready_for_query is used to indicate that the client process needs to be unlinked from the direct connection. In session mode, it is always linked

PR #456, opened 2024-10-04
fix: add local proxy multiplier

abc3:


test/supavisor/prom_ex_test.exs
```diff
@@ -31,7 +31,7 @@ defmodule Supavisor.PromExTest do
     :ok = GenServer.stop(proxy)
     :ok = Supavisor.stop({{:single, @tenant}, user, :transaction, db_name, nil})
 
-    Process.sleep(1000)
+    Process.sleep(3000)
```
hauleth, line 34: We will need to rework it in future.

hauleth (approved):

test/supavisor/prom_ex_test.exs
abc3: yeah, 100%

PR #464, opened 2024-10-23
feat: periodically check auth_query creds

abc3:


lib/supavisor/secret_checker.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  use GenServer
+  require Logger
+
+  alias Supavisor.Helpers
+
+  @interval :timer.seconds(15)
+
+  def start_link(args) do
+    name = {:via, Registry, {Supavisor.Registry.Tenants, {:secret_checker, args.id}}}
```
hauleth, line 12: Do we need that registry at all? It doesn't seem to be used for anything.

lib/supavisor/secret_checker.ex
```diff
@@ -0,0 +1,117 @@
...
+
+    GenServer.start_link(__MODULE__, args, name: name)
+  end
+
+  def init(args) do
+    Logger.debug("SecretChecker: Starting secret checker")
+    tenant = Supavisor.tenant(args.id)
+
+    [%{auth: auth, user: user} | _] =
+      Enum.filter(args.replicas, fn e -> e.replica_type == :write end)
```
hauleth, line 22: `Enum.find/2` wouldn't be better there? In the end we are interested only in first entry.

lib/supavisor/secret_checker.ex
```diff
@@ -0,0 +1,117 @@
...
+    {:noreply, %{state | check_ref: check()}}
+  end
+
+  def handle_info(msg, state) do
+    Logger.error("Unexpected message: #{inspect(msg)}")
+    {:noreply, state}
+  end
+
+  def terminate(_, state) do
+    :gen_statem.stop(state.conn)
```
hauleth, line 84: Weird that `Postgrex` do not offer `Postgrex.stop/1` function. Maybe it would be worth adding such function upstream?

lib/supavisor/secret_checker.ex
```diff
@@ -0,0 +1,117 @@
...
+    case Helpers.get_user_secret(conn, auth.auth_query, user) do
+      {:ok, secret} ->
+        method = if secret.digest == :md5, do: :auth_query_md5, else: :auth_query
+        secrets = Map.put(secret, :alias, auth.alias)
+        value = {:ok, {method, fn -> secrets end}}
+
+        update_cache =
+          case Cachex.get(Supavisor.Cache, state.key) do
+            {:ok, {:cached, {_, {old_method, old_secrets}}}} ->
+              method != old_method or not match?(^secrets, old_secrets.())
```
hauleth, line 101:
```suggestion
              method != old_method or secrets != old_secrets.()
```

lib/supavisor/secret_checker.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  def check(interval \\ @interval),
+    do: Process.send_after(self(), :check, interval)
+
+  def check_secrets(%{auth: auth, user: user, conn: conn} = state) do
+    case Helpers.get_user_secret(conn, auth.auth_query, user) do
+      {:ok, secret} ->
+        method = if secret.digest == :md5, do: :auth_query_md5, else: :auth_query
+        secrets = Map.put(secret, :alias, auth.alias)
+        value = {:ok, {method, fn -> secrets end}}
```
hauleth, line 96: Why not create it in `if`? It is not used anywhere else beyond that `if update_cache do …`.

lib/supavisor/secret_checker.ex
abc3: yep, it can be very useful during debugging

lib/supavisor/secret_checker.ex
abc3: yeah, I've been wanting to add this PR to Postgrex for a long time

lib/supavisor/secret_checker.ex
abc3: to stop looping when the first e.replica_type == :write is found?

hauleth (approved):

lib/supavisor/secret_checker.ex
hauleth: I am not 100% sure if it is good approach. Maybe we could use [`Process.set_label/1`](https://hexdocs.pm/elixir/Process.html#set_label/1) in the future instead?

PR #466, opened 2024-10-30
chore: update poolboy to v0.0.3 and enhance error logging

abc3:


lib/supavisor/client_handler.ex
```diff
@@ -194,7 +194,7 @@ defmodule Supavisor.ClientHandler do
            {:next_event, :internal, event}}
         else
           reason = "Invalid format for user or db_name"
-          Logger.error("ClientHandler: #{inspect(reason)}")
+          Logger.error("ClientHandler: #{inspect(reason)} #{inspect({user, db_name})}")
```
hauleth, line 197: We need to start using structured logging…

hauleth (approved):

lib/supavisor/client_handler.ex
abc3: we have the username and database name in the log metadata, but combining them here into a single string would make it more clear

chasers: Also put them in the metadata

PR #467, opened 2024-10-31
fix: update db_pid structure in terminate function

abc3:


mix.exs
```diff
@@ -68,7 +68,7 @@ defmodule Supavisor.MixProject do
 
       # pooller
       # {:poolboy, "~> 1.5.2"},
-      {:poolboy, git: "https://github.com/abc3/poolboy.git", tag: "v0.0.3"},
+      {:poolboy, git: "https://github.com/abc3/poolboy.git", tag: "v0.0.2"},
```
hauleth, line 71: Rollback?

hauleth (approved):

mix.exs
abc3: yeah

PR #471, opened 2024-11-06
fix: better handling of resetting active_counter

abc3:


hauleth (approved):

hauleth:
Something bonkers in the tests is happening.

PR #474, opened 2024-11-09
test: mark PromEx test as flaky

hauleth:
That tests are constantly causing us problems in CI, but as it is test for internal functionality that do not affect users directly. So to streamline testing disable running these tests in default run. To run them one can use

    mix test --include flaky


abc3 (approved):

hauleth:
I have no idea why these tests fail…

PR #475, opened 2024-11-09
fix: mark `:db_termination` as shutdown

hauleth:
That will reduce amount of error logs related to unexpected process shutdown (as `db_termination` is rather expected shutdown).


abc3:
But this won't terminate the linked client handler?

abc3 (changes requested):
But this won't terminate the linked client handler?

hauleth:
@abc3 everything should work as expected there, as it only changes behaviour in the supervisor. Exit message is still sent and should be handled as is.

abc3 (approved):

PR #479, opened 2024-11-12
test: add integration tests against Postgres.js

hauleth:

Add integration tests against Postgres.js Node library.


flake.nix
```diff
@@ -78,20 +83,31 @@
...
+                package = pkgs.postgresql_15;
                 initialScript = ''
                   ${builtins.readFile ./dev/postgres/00-setup.sql}
 
                   CREATE USER postgres SUPERUSER PASSWORD 'postgres';
                 '';
                 listen_addresses = "127.0.0.1";
                 port = 6432;
+                settings = {
+                  max_prepared_transactions = 262143;
```
hauleth, line 95: It is maximal possible value, in tests I have set it to `2000` as it doesn't matter really as long as it is greater than `0` I think.

.github/workflows/elixir.yml
```diff
@@ -127,6 +127,54 @@ jobs:
...
+    needs: [deps]
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Elixir
+        id: beam
+        uses: erlef/setup-beam@v1
+        with:
+          otp-version: '25.3.2.7'
+          elixir-version: '1.14.5'
```
abc3, line 142: wondering if we want to put these versions in the env vars

hauleth: Could be useful, but as a separate PR. There I simply copied what we have in other steps.

test/integration/js/postgres/copy.csv
```diff
@@ -0,0 +1,2 @@
+1	2	3
+4	5	6
```
abc3, line 2: what does it do?

hauleth: No idea, it is copied as is from `postgres.js` test folder. I haven't really checked if that is used anywhere in test suite.

abc3: let's try to run it without this file :D

hauleth: I have found that it is used in one test:
```js
t('Copy from file', async() => {
  await sql`create table test (x int, y int, z int)`
  await new Promise(async r => fs
    .createReadStream(rel('copy.csv'))
    .pipe(await sql`copy test from stdin`.writable())
    .on('finish', r)
  )

  return [
    JSON.stringify(await sql`select * from test`),
    '[{"x":1,"y":2,"z":3},{"x":4,"y":5,"z":6}]',
    await sql`drop table test`
  ]
})
```

PR #484, opened 2024-11-14
chore: remove unneeded Plugs

hauleth:
These plugs weren't used at all, as we do not serve any static assets nor we
need live reload for anything (as we do not have any form of web UI).


abc3 (approved):

hauleth:
I think that in future we should move `phoenix_live_view` to development dependencies as well, as we do not use it in production. This would reduce potential attack surface.

PR #493, opened 2024-11-21
feat: use microseconds for measuring pool checkouts

hauleth:

Currently we measure pools checkouts in milliseconds, but these do not give us enough resolution and almost everything is crammed into `le="1"` bucket.

## What is the new behavior?

Change time unit used for measures to microsecond so all measurements are multiplied by factor of `1000` which should give us better insights into real distribution.


abc3 (approved):

chasers (approved):

hauleth:
@chasers I have changed the metric name in Prometheus/VictoriaMetrics to include the unit, so it wouldn't be confused with "old" metrics, but it will require updates in the Grafana dashboards.

chasers (approved):

PR #497, opened 2024-11-25
fix: cleanly shutdown `DbHandler` on failure

hauleth:


lib/supavisor/client_handler.ex
```diff
@@ -470,7 +470,20 @@ defmodule Supavisor.ClientHandler do
...
+      try do
+        DbHandler.checkout(db_pid, data.sock)
+      rescue
+        err ->
+          Logger.error(
+            "ClientHandler: Error while connecting to the DB - #{Exception.message(err)}"
+          )
+
+          _ = DbHandler.stop(db_pid)
+          reraise err, __STACKTRACE__
```
abc3, line 484: does it print creds?

hauleth: It uses "regular" error handling from Erlang, so it will use "default" view as any other `raise` that is happening in the system.

hauleth: @abc3 so in fact it changes *nothing* about the error reporting. Because without that `try` there if some function call would raise, then that error would happen in exactly the same way as it is happening with this code.

haiderRizvii (approved):

hauleth:
It doesn't seem to be meaningful. So closing it right now.

PR #498, opened 2024-11-26
fix: always monitor the manager

chasers:
- proxy ClientHandler should terminate when manager terminates
- needs tests

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -437,8 +437,10 @@ defmodule Supavisor.ClientHandler do
                 upstream_verify: nil
               })
 
-            {:keep_state, %{data | auth: auth, manager: manager_ref},
-             {:next_event, :internal, :connect_db}}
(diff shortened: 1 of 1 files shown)


lib/supavisor/client_handler.ex
```diff
@@ -623,7 +629,8 @@ defmodule Supavisor.ClientHandler do
   end
 
   # pool's manager went down
-  def handle_event(:info, {:DOWN, _, _, _, reason}, state, data) do
+  def handle_event(:info, {:DOWN, ref, _, _, reason}, state, %{manager: manager_ref} = data)
+      when ref == manager_ref do
```
hauleth, line 633: I am fan of

```suggestion
  def handle_event(:info, {:DOWN, ref, _, _, reason}, state, %{manager: ref} = data) do
```

But I know that some people do not agree with me there.

hauleth (approved):

PR #531, opened 2024-12-29
test: check network usage reporting

hauleth:
Check if the network usage stats (main billing source) are properly reported.

### DIFF
--- a/test/supavisor/client_handler/stats_test.exs
+++ b/test/supavisor/client_handler/stats_test.exs
@@ -1,0 +1,179 @@
+defmodule Supavisor.ClientHandler.StatsTest do
+  use Supavisor.DataCase, async: false
+
+  @moduletag telemetry: true
+
+  setup :external_id
+
+  setup ctx do
+    :telemetry.attach(
+      {ctx.test, :client},
+      [:supavisor, :client, :network, :stat],
+      &__MODULE__.handle_event/4,
+      self()
+    )
+
+    :telemetry.attach(
+      {ctx.test, :db},
+      [:supavisor, :db, :network, :stat],
+      &__MODULE__.handle_event/4,
+      self()
+    )
+
+    on_exit(fn ->
+      :telemetry.detach({ctx.test, :client})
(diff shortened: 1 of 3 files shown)


test/supavisor/client_handler/stats_test.exs
```diff
@@ -0,0 +1,133 @@
...
+
+    # XXX: This should fail or not?
+    # test "increase on just auth", ctx do
+    #   external_id = ctx.external_id
+    #
+    #   assert_received {:telemetry,
+    #                    {:client, %{recv_oct: recv, send_oct: sent}, %{tenant: ^external_id}}}
+    #   assert recv > 0
+    #   assert sent > 0
+    # end
```
hauleth, line 70: I am not sure if this should be taken into account. cc @chasers @abc3

test/supavisor/client_handler/stats_test.exs
```diff
@@ -0,0 +1,133 @@
...
+          id: :postgrex_another
+        )
+
+      other_id = other.external_id
+
+      assert {:ok, _} = Postgrex.query(other_conn, "SELECT 1", [])
+
+      # assert_received {:telemetry, {:client, _, %{tenant: ^other_id}}}
+      refute_received {:telemetry, {:client, _, %{tenant: ^external_id}}}
+    end
```
hauleth, line 95: I am investigating failure of this test. It checks if querying data for one connection do not cause socket usage for another connection. But for whatever reason, it fails (which mean we have cross-reporting, or I have wrong assumptions about when metrics are sent).

chasers: Maybe metrics for external_id is reporting, but it's reporting zero?

hauleth: I will check that out. But it is weird that there is anything sent at all.

hauleth: Found the problem. It was reporting the data for handshake there, I needed to clear the message queue before doing anything extra.

chasers (approved):

PR #547, opened 2025-01-17
fix: validation prisma names

abc3:


hauleth (approved):

lib/supavisor/helpers.ex
```diff
@@ -371,9 +371,9 @@ defmodule Supavisor.Helpers do
 
   @spec validate_name(String.t()) :: boolean()
   def validate_name(name) do
-    # 1-63 characters, starting with a lowercase letter or underscore, and containing only alphanumeric characters, underscores, and dollar signs. Names with spaces or uppercase letters must be enclosed in double quotes.
+    # 1-63 characters, starting with a lowercase letter or underscore, and containing only alphanumeric characters, underscores, dollar signs, and hyphens. Names with spaces or uppercase letters must be enclosed in double quotes.
     String.length(name) <= 63 and
-      name =~ ~r/^(?:[a-z_][a-z0-9_$ ]*|"[a-zA-Z0-9_$ ]+")$/ and
+      name =~ ~r/^(?:[a-z_][a-z0-9_$\-]*|"[a-zA-Z0-9_$\-]+")$/ and
```
hauleth, line 376: Technically that escape isn't needed there.

lib/supavisor/helpers.ex
```diff
@@ -371,9 +371,9 @@ defmodule Supavisor.Helpers do
 
   @spec validate_name(String.t()) :: boolean()
   def validate_name(name) do
-    # 1-63 characters, starting with a lowercase letter or underscore, and containing only alphanumeric characters, underscores, and dollar signs. Names with spaces or uppercase letters must be enclosed in double quotes.
+    # 1-63 characters, starting with a lowercase letter or underscore, and containing only alphanumeric characters, underscores, dollar signs, and hyphens. Names with spaces or uppercase letters must be enclosed in double quotes.
     String.length(name) <= 63 and
```
hauleth, line 375: Technically we could use `byte_size/1` there, as later we check if the name is ASCII only, so we do not need `String.length/1` there (which is Unicode aware).

hauleth (approved):

lib/supavisor/helpers.ex
abc3: well, it needs
```
== Compilation error in file lib/supavisor/helpers.ex ==
** (Regex.CompileError) range out of order in character class at position 20
    (elixir 1.17.1) lib/regex.ex:257: Regex.compile!/2
    (elixir 1.17.1) expanding macro: Kernel.sigil_r/2
    lib/supavisor/helpers.ex:376: Supavisor.Helpers.validate_name/1
    (elixir 1.17.1) expanding macro: Kernel.and/2
    lib/supavisor/helpers.ex:375: Supavisor.Helpers.validate_name/1
    (elixir 1.17.1) expanding macro: Kernel.and/2
    lib/supavisor/helpers.ex:376: Supavisor.Helpers.validate_name/1
```

hauleth: Weird, I was sure it worked for me. If it is needed, then it is needed, and we cannot do anything about it 👍

PR #549, opened 2025-01-17
chore: remove check for version updated on each PR

hauleth:


chasers (changes requested):
@hauleth why is this important right now?

hauleth:
@chasers it is not important right now. It is just opened for the future. Should I close it and open it later?

hauleth:
I closed, but recently it was really painful to keep that file up to date. Especially with sheer amount of little changes, especially one that aren't really related to application code.

PR #560, opened 2025-01-21
test: integration test for stats on other nodes

hauleth:


hauleth:
It seems that there is some interference from other tests. It works when ran on its own, but fails when ran with others. I will investigate it.

chasers (approved):

PR #574, opened 2025-01-24
feat: Allow infinite reconnect retries with -1 setting

kahirokunn:

This PR introduces a feature that enables Supavisor to allow infinite reconnect retries when @reconnect_retries is set to -1.

Currently, the reconnect logic has a hard limit. Once the number of retries exceeds @reconnect_retries, Supavisor stops attempting to reconnect if the database is unavailable.

## What is the new behavior?

With this change, if @reconnect_retries is set to -1, Supavisor will never stop retrying to connect to the database. 
## Additional context

This supports environments that use dynamic scaling or “scale to zero” strategies where the database may be intentionally down but needs to be brought online seamlessly when connections are attempted again.
Makes “infinite” reconnection behavior a conscious user choice, helping in scenarios where the DB is scaled down and then scaled back up based on metrics.

### DIFF
--- a/lib/supavisor/db_handler.ex
+++ b/lib/supavisor/db_handler.ex
@@ -110,7 +110,7 @@ defmodule Supavisor.DbHandler do
       ]
 
     maybe_reconnect_callback = fn reason ->
-      if data.reconnect_retries > @reconnect_retries and data.client_sock != nil,
+      if @reconnect_retries != -1 and data.reconnect_retries > @reconnect_retries and data.client_sock != nil,
         do: {:stop, {:failed_to_connect, reason}},
         else: {:keep_state_and_data, {:state_timeout, reconnect_timeout(data), :connect}}
     end


hauleth (commented):
I wonder how is that useful vs setting reconnection count to some large value. Because TBH I so not see how that can be safe. It is compile-time only option, so it will require some footwork to apply anyway, but I simply cannot imagine situation when you want to have really unlimited amount of retries.

kahirokunn:
I really want to SLEEP until I get a valid IP for the following EndpointSlice, but I thought it would be too much Kubernetes knowledge for the quirks.

https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#endpointslice-resource

However, I think it would be ideal if something like a plugin function could be implemented where we could write this kind of environment-dependent code and retry logic.

v0idpwn:
Hi, @kahirokunn. Sorry for the long lag. I opened #705, which allows changing this value without rebuilding, I'll close this PR in favor of it.

kahirokunn:
This is exactly what I was looking for!
Thank you very much!

PR #575, opened 2025-01-27
test: allow changing verbosity in captured logs

hauleth:
This is useful in some development sessions, as by default the printed logs can be quite lengthy, especially in integration tests. This change adds support for using `TEST_LOGS` environment variable to set the captured verbosity to values that make it easier to debug the output.


hauleth:
@chasers @abc3 I am thinking also about setting console logger in tests to `level: :none`, so it would clean output of the test suite, to print logs only on test failure.

abc3:
Sounds good for me

hauleth:
I have added that change as well.

chasers (approved):

PR #577, opened 2025-01-27
fix: lower log level when removing stopped listener

hauleth:
This message is not an error, and it needlessly scares operators. This error seems like it is bug in Ranch, as according to the docs there should be no error and instead we should get `{:error, :not_found}` response, but for now let's deal with it by lowering the error message to make it simpler.


chasers (changes requested):
If we're updating logs, needs a proper error code, and description like:

https://supabase.github.io/supavisor/monitoring/logs/

And tag the project in the metadata.

I know you were thinking about Postgres like error codes, but Realtime and Auth are doing these style so let's stick with them for now.

hauleth:
I would need to read more how it should look like there, as I am not sure if I understand your idea correctly. You mean that I should have some keyword marker in that log message so it would be easier to search?

chasers:
Yeah it's just a prefix eg https://github.com/supabase/supavisor/blob/7e16258464312ae9a223104c50f08d27c4c87e4c/lib/supavisor/client_handler.ex#L628

chasers:
@hauleth that is the module name not an error code.

Also need:
- [ ] docs updated
- [ ] better error message
- [ ] project metadata

hauleth:
@chasers now it looks ok?

chasers (approved):

PR #580, opened 2025-01-28
fix: more correct DB name handling

### LINKED ISSUE #579: Incorrect handling of the database and user name
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

There is DB and username validation that do not conform to what is supported by the Postgres.

## To Reproduce

Create database with name that contains at least one of:

- unquoted capital letter
- non-ASCII letter

Try to connect to such database.

## Expected behavior

It should be possible to connect to that DB.

hauleth:
PostgreSQL requires that the DB name is a valid identifier

https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

This was causing some issues with metrics gathering in the past, but
that should be resolved now. This makes it more in-line with what
PostgreSQL is allowing and expecting. The max length is compile-time
configurable, as PostgreSQL technically provides a way to increase the
byte limit, by default we use current PostgreSQL default there.

This mean that some users *may* be allowed to use weird names, but
assuming that Victoria Metrics is properly parsing exported metrics
using OpenMetrics specs, then it should work there.

Close #579


abc3 (approved):

abc3 (approved):

abc3:
* db_name and **user_name**

PR #586, opened 2025-01-30
feat: use iodata for metrics export

hauleth:
This prevents memory bloat form creating all intermediate binaries that are later dropped. Instead we try to reuse as much of binaries as possible and then let the VM do the concatenation if needed.


abc3 (approved):
it would be nice to add more tests to ensure it works properly

filipecabaco (approved):

PR #599, opened 2025-02-19
fix: set sync_threshold and discard_threshold

chasers:
Sets these such that Supavisor will drop logs instead of going into sync mode.


config/prod.exs
```diff
@@ -14,6 +14,8 @@ import Config
 # Do not print debug messages in production
 config :logger,
   level: :info,
+  sync_threshold: 10_000,
+  discard_threshold: 10_000,
```
hauleth, line 18: I need to check if that still works like that, as it is no longer mentioned by the `Logger` docs. I think we need to add `:logger_backends` to make it work properly.

hauleth (approved):

PR #600, opened 2025-02-19
chore: add .tool-versions

chasers:
- adds .tool-verisons I am not smart enough for nix


.tool-versions
```diff
@@ -0,0 +1,4 @@
+rust 1.84.1
+elixir 1.18.2-otp-27
+nodejs 18.13.0
+erlang 27.2
```
hauleth, line 4: Weird that it doesn't add NL at end by default.

hauleth (approved):

PR #603, opened 2025-02-21
fix: do not verify TLS cert when connecting to the DB

hauleth:


hauleth:
To clarify, it **is** safe, as it is how it was working in OTP 25. OTP 26 or 27 changed that to default to `:verify_peer`. This forces old behaviour (as otherwise it will fail). I wasn't sure if I can use OS provided root certs to verify that certificate, so I set it to `:verify_none`. This still uses secure communication.

filipecabaco (approved):

PR #610, opened 2025-02-24
fix: correctly cleanup metrics tables

hauleth:


filipecabaco (approved):
out of curiosity why handle continue vs handle info ?

hauleth:
`handle_info` calls `handle_continue`, and the reason for that is to deduplicate code.

PR #613, opened 2025-02-25
feat: add `instance_id` to metrics metadata

hauleth:


lib/supavisor/monitoring/prom_ex.ex
```diff
@@ -30,12 +30,7 @@ defmodule Supavisor.Monitoring.PromEx do
 
   @spec set_metrics_tags() :: map()
   def set_metrics_tags do
-    [_, host] = node() |> Atom.to_string() |> String.split("@")
-
-    metrics_tags = %{
-      region: Application.fetch_env!(:supavisor, :region),
-      host: host
-    }
+    metrics_tags = :logger.get_primary_config()[:metadata] || %{}
```
hauleth, line 33: In theory that is not needed as `metadata` there should always be map.

chasers (approved):

PR #616, opened 2025-02-26
fix: do not handle errors during connection

hauleth:
This is potential source of a problem, as if there are is error in 2nd or 3rd clause of `with` then we are leaking connections which in the end may result in the connections exhaustion. Instead we should fail loudly that there is something wrong with the connections. It should also help with debugging potential problems.

### DIFF
--- a/lib/cluster/strategy/postgres.ex
+++ b/lib/cluster/strategy/postgres.ex
@@ -52,24 +52,19 @@ defmodule Cluster.Strategy.Postgres do
   end
 
   def handle_continue(:connect, state) do
-    with {:ok, conn} <- Postgrex.start_link(state.meta.opts.()),
-         {:ok, conn_notif} <- Postgrex.Notifications.start_link(state.meta.opts.()),
-         {_, _} <- Postgrex.Notifications.listen(conn_notif, state.config[:channel_name]) do
-      Logger.info(state.topology, "Connected to Postgres database")
+    {:ok, conn} = Postgrex.start_link(state.meta.opts.())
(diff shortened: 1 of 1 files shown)


lib/cluster/strategy/postgres.ex
```diff
@@ -52,24 +52,19 @@ defmodule Cluster.Strategy.Postgres do
...
-        state.meta
-        | conn: conn,
-          conn_notif: conn_notif,
-          heartbeat_ref: heartbeat(0)
-      }
-
-      {:noreply, put_in(state.meta, meta)}
-    else
-      reason ->
-        Logger.error(state.topology, "Failed to connect to Postgres: #{inspect(reason)}")
```
chasers, line 70: What happens when this fails loudly does it take down the whole app with it?

hauleth: I would need to check exactly, but in theory, if the failure happen repeatedly, then it can cause whole application to go down.

abc3: > does it take down the whole app with it?


No, Supervisor will try to restart the Postgres strategy process, and after exhausting the retries, it will just leave it unstarted. But that change doesn’t make much sense anyway, because the node will be out of the cluster in case of a Postgres failure. You probably want to forward already assigned values, send a connect message after a timeout, and handle them more gracefully

abc3:
> leaking connections

In case of failure in the following cases, the node will not handle notifications because there is no retry logic. So yeah, there’s a potential place for a bug, but it doesn’t lead to leaking connections

PR #618, opened 2025-03-02
fix: remove duplicate config entry

hkrutzer:
Copy of this line was introduced in afc0995
but it was already present.


hauleth (approved):

hauleth:
Thanks, I have missed it. Could you also bump `VERSION`?

hkrutzer:
Yes, done

PR #623, opened 2025-03-05
chore: add mix_audit

hauleth:


.github/workflows/elixir.yml
```diff
@@ -61,6 +61,29 @@ jobs:
...
+          elixir-version: '1.18.2'
+      - name: Cache Mix
+        uses: actions/cache@v4
+        with:
+          path: deps
+          key: ${{ runner.os }}-mix-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+      - name: Compile deps
+        run: mix deps.compile
+      - name: mix_audit check
+        run: mix deps.audit
```
hauleth, line 85: I think that it should be moved to separate action file and should probably be ran on schedule instead of being ran on PR.

filipecabaco (approved):

PR #624, opened 2025-03-06
test: use default time limit on assert_receive in stats tests

hauleth:


filipecabaco (approved):

test/supavisor/client_handler/stats_test.exs
```diff
@@ -127,7 +127,7 @@ defmodule Supavisor.ClientHandler.StatsTest do
       this = Node.self()
 
       refute_receive {^telemetry, {:client, _, %{tenant: "proxy_tenant1"}}, ^node}
-      assert_receive {^telemetry, {:client, _, %{tenant: "proxy_tenant1"}}, ^this}, 1000
+      assert_receive {^telemetry, {:client, _, %{tenant: "proxy_tenant1"}}, ^this}
```
hauleth, line 130: Ehhh, seems that default timeout is 100ms which is even worse. I need to bump it up it seems. It is very flaky test and I cannot get why it is happening in CI but not locally.

filipecabaco: yep the default is 100ms I actually thought that you wanted to find a way to speed this one up 😅

hauleth: @filipecabaco seems that 10s timeout works ok :D

PR #628, opened 2025-03-07
test: create separate tenant for stats tests instead of reusing existing one

hauleth:


lib/supavisor/helpers.ex
```diff
@@ -89,7 +89,10 @@ defmodule Supavisor.Helpers do
     end
   end
 
-  @spec get_user_secret(pid(), String.t(), String.t()) :: {:ok, map()} | {:error, String.t()}
+  @spec get_user_secret(pid(), String.t() | nil, String.t()) ::
+          {:ok, map()} | {:error, String.t()}
+  def get_user_secret(_conn, nil, _user), do: {:error, "No auth_query specified"}
```
hauleth, line 94: I am not 100% sure how to handle such case and I do not know why sometimes it is `nil` in tests and sometimes it is not.

chasers: Hmm auth query is not in the tenant record.

hauleth: Yes, and I have checked in the schema, it is not required (at least in code). I do not know if that is bug there or in schema.

hauleth:
I hate that single test. I cannot reproduce the problem locally and it is flaky as hell…

PR #633, opened 2025-03-26
Remove required fields from OpenAPI schema that don’t exist

plarson:

Bug fix - Corrects the OpenAPI schema definition by removing invalid required field specifications.

I used `swagger2openapi` to generate an OpenAPI yml from the swagger endpoint. The OpenAPI schema in `lib/supavisor_web/open_api_schemas.ex` incorrectly lists some fields as required that don't actually exist in the schema definition, which makes client side OpenAPI validation fail. 

## What is the new behavior?

The schema's required fields list has been updated to only include fields that actually exist in the schema definition.


v0idpwn (approved):

v0idpwn:
Thank you! 💚

PR #654, opened 2025-06-09
feat: add ipv6 support for supavisor internal database

### LINKED ISSUE #649: Supavisor cannot resolve IPV6
# Bug report

## Describe the bug
Using the self-hosted Supabase template on Railway : https://railway.com/template/supabase.

After adding this Supavisor repository to Railway: https://github.com/supabase/supavisor (since it is not included in the template above)

Supavisor cannot resolve IPV6 domains, preventing me from using Railway's internal network. 

## To Reproduce
I did set expected variables in the Supavisor Container: 

ECTO_IPV6=true
API_JWT_SECRET
METRICS_JWT_SECRET
VAULT_ENC_KEY
SECRET_KEY_BASE
DATABASE_URL=ecto://${DB_USER}:${DB_PASSWORD}@${DB_HOST}:5432/${DB_NAME}

DB_HOST being Railway's internal network address for my PostgreSQL db.

Error raised by Supavisor: 
12:52:10.438 [error] Postgrex.Protocol (#PID<0.3326.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (postgres-ts0p.railway.internal:5432): non-existing domain - :nxdomain

## Expected behavior

I want Supavisor to connect to my database using Railway's internal network (thus use IPV6)

abc3:
This PR introduces IPv6 support for Supavisor's internal metadata database. Users can enable IPv6 by setting the `DB_IP_VERSION` environment variable to "ipv6". By default, connections will continue to use IPv4. 

Closes #649


config/runtime.exs
```diff
@@ -143,6 +143,11 @@ downstream_key =
     end
   end
 
+db_socket_options =
+  if System.get_env("DB_IP_VERSION") == "ipv6",
+    do: [:inet6],
+    else: [:inet]
```
hauleth: I wonder - if we will default to IPv6 will it work with IPv4 sockets as well?

config/runtime.exs
```diff
@@ -143,6 +143,11 @@ downstream_key =
     end
   end
 
+db_socket_options =
+  if System.get_env("DB_IP_VERSION") == "ipv6",
```
hauleth: I would prefer to start prefixing all environment variables with `SUPAVISOR_` to explicitly describe why these are needed.

abc3: That's a good call, I added the same variable as in Realtime, but prefixing it with SUPAVISOR_ makes more sense

config/runtime.exs
abc3: I don't have such an environment right now, but when I tried it a few months ago, it didn't work

config/runtime.exs
abc3: added prefix

Issue #663, opened 2025-06-16
Parameter server_version changed" error after PostgreSQL upgrade despite Supavisor restart


nitindhiman314e:
Hi Team,

After upgrading PostgreSQL from version 16.3 to 16.9, Supavisor (v2.0.21) is throwing the following error:

`[error] Parameter status error: "Parameter server_version changed from 16.3 to 16.9"
`

This happens even after:

- Restarting both the PostgreSQL server and Supavisor.
- Deleting Supavisor pods to force a clean restart.
- Ensuring there are no stale Postgres connections (verified via pg_stat_activity).

The connections seen in PostgreSQL are new and initiated after the upgrade, yet Supavisor appears to expect the old version (16.3) when negotiating the connection.

Please suggest how we can mitigate this issue. 

Thanks.

v0idpwn:
Hi, @nitindhiman314e, thanks for the report. Supavisor is supposed to handle this, and update the version internally right after the error log you sent. I'll be trying to reproduce and see how it behaves.

A potential mitigation is updating the tenants manually. or re-creating the tenants in the API. Could you please try that?

v0idpwn:
Hi again, @nitindhiman314e. I just tested doing an update. The first connection attempt crashed, but it performed the parameter update as a side effect and next connections succeed.

Here are the logs from my attempt (notice that I have debug logs enabled, you may not have them):
```
17:02:39.151 project=t2_tenant user=postgres region=eu type=single [error] Parameter status error: "Parameter server_version changed from 14.6 (Debian 14.6-1.pgdg110+1) to 15.8"

17:02:39.153 project=t2_tenant user=postgres region=eu type=single [debug] QUERY OK source="tenants" db=1.8ms idle=20.2ms
SELECT t0."id", t0."db_host", t0."db_port", t0."db_database", t0."external_id", t0."default_parameter_status", t0."ip_version", t0."upstream_ssl", t0."upstream_verify", t0."upstream_tls_ca", t0."enforce_ssl", t0."require_user", t0."auth_query", t0."default_pool_size", t0."sni_hostname", t0."default_max_clients", t0."client_idle_timeout", t0."client_heartbeat_interval", t0."allow_list", t0."availability_zone", t0."inserted_at", t0."updated_at" FROM "_supavisor"."tenants" AS t0 WHERE (t0."external_id" = $1) ["t2_tenant"]
[...]

PR #666, opened 2025-06-17
fix: don't stop client connection when parameter status is updated

v0idpwn:
Just continue with the greeting process, with the new parameter status. Also adds a test script that performs
a pseudo postgres update. In the original code, the query must be retried on Supavisor because the first 
connection attempt fails. After the patch, no retries are needed.

Discovered in #663.

### DIFF
--- a/lib/supavisor/manager.ex
+++ b/lib/supavisor/manager.ex
@@ -93,9 +93,9 @@ defmodule Supavisor.Manager do
 
         {:error, reason} ->
           Logger.error("Parameter status error: #{inspect(reason)}")
-          new_ps = %{server_version: ps["server_version"]}
-          Tenants.update_tenant_ps(state.tenant, new_ps)
-          :updated
+          new_ps = %{"server_version" => ps["server_version"]}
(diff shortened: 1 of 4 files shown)


chasers (approved):

lib/supavisor/manager.ex
```diff
@@ -147,16 +135,24 @@ defmodule Supavisor.Manager do
...
-  defp check_parameter_status(ps, def_ps) do
-    Enum.find_value(ps, :ok, fn {key, new_value} ->
-      case def_ps do
-        %{^key => old_value} when old_value != new_value ->
-          {:error, "Parameter #{key} changed from #{old_value} to #{new_value}"}
-
-        _ ->
-          nil
-      end
+  @spec maybe_update_parameter_status(binary, map, map) :: {:ok, Tenants.Tenant.t()}
```
v0idpwn:
```suggestion
  @spec maybe_update_parameter_status(binary, map, map) :: :ok
```

abc3:
hey @v0idpwn, looks good to me!

PR #667, opened 2025-06-18
fix(logger): format mfas in accordance to logflare, put level in meta

v0idpwn:

Changes file log formatting to conform to Logflare's format.

Because MFAs have a mix of strings and integers, ingestion fails. Also, the log level is on top level, but logflare expects it to be inside metadata, so it wouldn't get picked up.

## What is the new behavior?

Added a pass to perform arbitrary transformations.

## Additional context

Tested in staging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ priv/native/*
 *.bggg
 /.pre-commit-config.yaml
 *.coverdata
+/tmp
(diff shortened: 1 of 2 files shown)


lib/supavisor/logger/logflare_formatter.ex
```diff
@@ -26,15 +27,27 @@ defmodule Supavisor.Logger.LogflareFormatter do
...
+    mfa: &__MODULE__.transform_mfa/1
+  ]
+
   @impl true
   def check_config(_), do: :ok
 
   @impl true
   def format(%{msg: msg, level: level, meta: meta}, opts) do
-    context_keys = [Map.get(opts, :context, []) | @default_context]
+    context_keys = Map.get(opts, :context, []) ++ @default_context
```
v0idpwn, line 39: This was also broken, the keys went outside of `context` because the list wasn't properly merged

PR #678, opened 2025-06-26
fix: race on tenant startup causing connection failure

v0idpwn:
Noticed this while starting many connections quickly for a tenant in local development/testing.


chasers (approved):

nassmim:
I have the same issue and it's been almost two weeks now I'm not able to connect.
Do we have some restart or any updates to do following your fix?

PR #688, opened 2025-06-30
feat: add metric for client connection lifetime

v0idpwn:
Store start_time in client connection registry, then periodically emit telemetry for active client connections.

### DIFF
--- a/lib/supavisor/monitoring/tenant.ex
+++ b/lib/supavisor/monitoring/tenant.ex
@@ -40,18 +40,31 @@ defmodule Supavisor.PromEx.Plugins.Tenant do
 
     use Peep.Buckets.Custom,
       buckets: [
+        # 0.5 seconds
         500,
+        # 1 second
         1_000,
+        # 5 seconds
         5_000,
-        10_000,
+        # 20 seconds
+        20_000,
+        # 1 minute
         60_000,
+        # 5 minutes
         300_000,
+        # 30 minutes
         1_800_000,
+        # 2 hours
         7_200_000,
+        # 8 hours
         28_800_000,
+        # 1 day
         86_400_000,
+        # 3 days
         259_200_000,
+        # 1 week
         604_800_000,
+        # 30 days
         2_592_000_000
       ]
   end
@@ -251,7 +264,7 @@ defmodule Supavisor.PromEx.Plugins.Tenant do
       {__MODULE__, :execute_client_connections_lifetime, []},
       [
         distribution(
-          [:supavisor, :client, :connection, :lifetime],
+          [:supavisor, :client, :connection, :lifetime, :ms],
           event_name: [:supavisor, :client, :connection, :lifetime],
           measurement: :lifetime,
           description: "How long the client connection has been alive.",
@@ -265,6 +278,7 @@ defmodule Supavisor.PromEx.Plugins.Tenant do
     )
   end
 
+  @spec execute_client_connections_lifetime() :: :ok
   def execute_client_connections_lifetime do
     read_time = System.monotonic_time()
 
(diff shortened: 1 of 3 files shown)


lib/supavisor/monitoring/tenant.ex
```diff
@@ -34,6 +35,25 @@ defmodule Supavisor.PromEx.Plugins.Tenant do
...
+        10_000,
+        60_000,
+        300_000,
+        1_800_000,
+        7_200_000,
+        28_800_000,
+        86_400_000,
+        259_200_000,
+        604_800_000,
+        2_592_000_000
```
v0idpwn: I wanted to use more "semantic" times here with e.g.: `timer.hours(24)` or `24 * 3600 * 1000`, but it turns out that Peep does some ast manipulation and can only deal with integer literals. Can add comments instead but looks pretty ugly :)

lib/supavisor/monitoring/tenant.ex
```diff
@@ -222,6 +244,60 @@ defmodule Supavisor.PromEx.Plugins.Tenant do
...
   end
 
+  defp client_connections_lifetime(poll_rate) do
+    Polling.build(
+      :supavisor_client_connections_lifetime,
+      poll_rate,
+      {__MODULE__, :execute_client_connections_lifetime, []},
+      [
+        distribution(
+          [:supavisor, :client, :connection, :lifetime],
```
abc3: adding units to metrics can make them easier to understand when working in grafana

```suggestion
          [:supavisor, :client, :connection, :lifetime, :ms],
```

lib/supavisor/monitoring/tenant.ex
```diff
@@ -222,6 +244,60 @@ defmodule Supavisor.PromEx.Plugins.Tenant do
...
+
+  def execute_client_connections_lifetime do
+    read_time = System.monotonic_time()
+
+    Supavisor.Registry.TenantClients
+    |> Registry.select([{{:"$1", :_, :"$2"}, [], [{{:"$1", :"$2"}}]}])
+    |> Enum.each(&emit_client_connection_lifetime(&1, read_time))
+  end
+
+  def emit_client_connection_lifetime(
```
abc3, line 291:
```suggestion
  @spec emit_client_connection_lifetime(Supavisor.id(), integer()) :: :ok | :noop
  def emit_client_connection_lifetime(
```

lib/supavisor/monitoring/tenant.ex
abc3: Yeah, under the hood, Peep does some macro magic, and it expects a plain list of numbers when you define custom buckets. So the most reliable and readable way is to just use comments

v0idpwn: What's sad is that the formatter won't let me do 

```elixir
[
  1000, # 1 second
  5000, # 5 seconds
]
```

as it formats to
```elixir
[
# 1 second
1000,
# 5 seconds
5000
]
```

😭

abc3: made an issue just in case https://github.com/rkallos/peep/issues/44

lib/supavisor/monitoring/tenant.ex
```diff
@@ -222,6 +257,62 @@ defmodule Supavisor.PromEx.Plugins.Tenant do
...
+  @spec execute_client_connections_lifetime() :: :ok
+  def execute_client_connections_lifetime do
+    read_time = System.monotonic_time()
+
+    Supavisor.Registry.TenantClients
+    |> Registry.select([{{:"$1", :_, :"$2"}, [], [{{:"$1", :"$2"}}]}])
+    |> Enum.each(&emit_client_connection_lifetime(&1, read_time))
+  end
+
+  @spec emit_client_connection_lifetime(Supavisor.id(), integer()) :: :ok | :noop
```
abc3: missed a keyword 
```suggestion
  @spec emit_client_connection_lifetime({Supavisor.id(), keyword()}, integer()) :: :ok | :noop
```

PR #689, opened 2025-07-01
feat: monitor large process heaps and long message queues

abc3:
This PR introduces monitoring for large process heaps in `Supavisor.ErlSysMon` using the `:large_heap` system flag. The threshold is set to `25MB` and is computed using the new helper function `Supavisor.Helpers.mb_to_words/1`.

It enables detection of processes that grow unusually large after garbage collection, helping to catch potential memory issues early.

Test coverage for `Supavisor.ErlSysMon` has also been added.

Updated:
- Added monitoring for long message queues using the :long_message_queue system flag:
```elixir
{:long_message_queue, {0, 1_000}}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 # Supavisor - Postgres connection pooler
 
+[![Coverage Status](https://coveralls.io/repos/github/supabase/supavisor/badge.svg?branch=main)](https://coveralls.io/github/supabase/supavisor?branch=main)
+
 - [Overview](#overview)
 - [Motivation](#motivation)
 - [Architecture](#architecture)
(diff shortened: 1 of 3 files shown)


lib/supavisor/erl_sys_mon.ex
```diff
@@ -4,19 +4,22 @@ defmodule Supavisor.ErlSysMon do
...
   end
 
   def init(_args) do
     :erlang.system_monitor(self(), [
       :busy_dist_port,
       :busy_port,
       {:long_gc, 250},
-      {:long_schedule, 100}
+      {:long_schedule, 100},
+      {:large_heap, Helpers.mb_to_words(25)}
```
v0idpwn, line 23: Should we maybe add `long_message_queue` as well?

abc3: sure, will copy from realtime

v0idpwn (approved):

PR #691, opened 2025-07-01
feat: check database and erpc capabilities on health check

v0idpwn:
Introduces `Supavisor.Health`, which provides a function that runs health checks.

Added two checks:
- Acceptable ERPC latencies: fails if a node has high latency to all other nodes through :erpc. Doesn't run if in a 1 or 2 node cluster. Fails if **all requests** have latency over 500ms or fail.
- Database reachable: fails if can't run a simple query in the database.

Calls this function on the health check endpoint, and return 503 if health checks are failing. After some time, if the condition persists, the infrastructure should restart the instance.

### DIFF
--- a/lib/supavisor/health.ex
+++ b/lib/supavisor/health.ex
@@ -73,17 +73,15 @@ defmodule Supavisor.Health do
   @doc false
   @spec database_reachable?() :: boolean()
   def database_reachable? do
-    try do
-      case Supavisor.Repo.query("SELECT 1") do
-        {:ok, %Postgrex.Result{rows: [[1]]}} ->
-          true
+    case Supavisor.Repo.query("SELECT 1") do
+      {:ok, %Postgrex.Result{rows: [[1]]}} ->
+        true
 
-        _ ->
-          false
-      end
-    catch
-      _, _ -> false
+      _ ->
+        false
     end
+  catch
+    _, _ -> false
   end
 
   @doc false
--- a/lib/supavisor/health.ex
+++ b/lib/supavisor/health.ex
@@ -60,10 +60,11 @@ defmodule Supavisor.Health do
         true
 
       nodes ->
-        # If **any** other node returns replies within the timeout, we are good.
         results = :erpc.multicall(nodes, fn -> :ok end, acceptable_erpc_latency)
 
-        Enum.zip(nodes, results)
+        # If **any** other node returns replies within the timeout, we are good.
+        nodes
+        |> Enum.zip(results)
         |> Enum.reduce(false, fn {node, result}, acc ->
           case result do
             {:ok, _} ->
--- a/lib/supavisor/health.ex
+++ b/lib/supavisor/health.ex
@@ -20,7 +20,7 @@ defmodule Supavisor.Health do
   @doc """
   The main API for checking the health of a Supavisor node.
   """
-  @spec health_check([Keyword.t()]) :: :ok | {:error, :failed_checks, [atom()]}
+  @spec health_check(Keyword.t()) :: :ok | {:error, :failed_checks, [atom()]}
   def health_check(checks \\ @checks) do
     successful_checks =
       @task_supervisor
--- a/lib/supavisor/health.ex
+++ b/lib/supavisor/health.ex
@@ -11,7 +11,7 @@ defmodule Supavisor.Health do
   require Logger
 
   @checks [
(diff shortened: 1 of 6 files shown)


lib/supavisor/health.ex
```diff
@@ -0,0 +1,93 @@
...
+      |> Task.Supervisor.async_stream_nolink(checks, fn {check, {m, f, a}} ->
+        result = apply(m, f, a)
+        {check, result}
+      end)
+      |> Enum.reduce([], fn result, acc ->
+        case result do
+          {:ok, {check, true}} -> [check | acc]
+          {:ok, {_check, false}} -> acc
+          {:exit, _} -> acc
+        end
```
v0idpwn, line 36: If we want to be **really defensive** here, we can make it consider crashing checks as successful (so that a poorly written check doesn't cause node restarts)

lib/supavisor/health.ex
```diff
@@ -0,0 +1,93 @@
...
+        true
+
+      nodes ->
+        # If **any** other node returns replies within 500ms, we are good.
+        Enum.any?(nodes, fn node ->
+          try do
+            :ok = :erpc.call(node, fn -> :ok end, 500)
+            true
+          catch
+            _, _ -> false
```
chasers: We’re going to want to know what this error is if it happens. Can we log something?

chasers (changes requested):

chasers:
Should run in a two node cluster also. Why not?

chasers:
Otherwise looks great

lib/supavisor/health.ex
```diff
@@ -0,0 +1,91 @@
...
+  """
+
+  require Logger
+
+  @checks [
+    acceptable_erpc_latencies: {__MODULE__, :acceptable_erpc_latencies?, []},
+    database_reachable: {__MODULE__, :database_reachable?, []}
+  ]
+
+  @task_supervisor __MODULE__.TaskSupervisor
```
chasers, line 18: Missing this in the application.ex I believe?

lib/supavisor/health.ex
chasers: Lets not write pooly written checks :D This is an important part here

lib/supavisor/health.ex
v0idpwn: I'm starting `Supavisor.Health` there, and the child_spec for `Supavisor.Health` specifies it (it's at the bottom of the module).

chasers: Ah ya nice!

chasers (approved):

lib/supavisor/health.ex
```diff
@@ -0,0 +1,91 @@
...
+      nodes ->
+        # If **any** other node returns replies within 500ms, we are good.
+        Enum.any?(nodes, fn node ->
+          try do
+            :ok = :erpc.call(node, fn -> :ok end, 500)
+            true
+          catch
+            _, _ -> false
+          end
+        end)
```
abc3: You could simplify this part using `:erpc.multicall/4` to perform the latency checks in parallel.
Also, instead of manually matching `[]` and `[_]`, you can just use pattern matching with `[_ | _]` to handle the 2+ nodes case, and fallback to `_ -> true` for 0 or 1 nodes

lib/supavisor_web/controllers/tenant_controller.ex
```diff
@@ -209,11 +209,18 @@ defmodule SupavisorWeb.TenantController do
...
   )
 
   def health(conn, _) do
-    send_resp(conn, 204, "")
+    case Supavisor.Health.health_check() do
+      :ok ->
+        send_resp(conn, 204, "")
+
+      {:error, :failed_checks, _failed_checks} ->
+        send_resp(conn, 503, "")
```
abc3: Wondering if it would make sense to include a reason in the 503 response. This could help with observability or debugging, especially when multiple health checks are involved

lib/supavisor/health.ex
v0idpwn: The other side of `:erpc.multicall` vs `Enum.any?/2` is that `any?/2` will avoid performing calls to the whole cluster if the first one succeeds 🤔 . But I think it's a good change nonetheless, will do.

Edit: specially now that we want to log failed ones, it's mandatory to call everyone

abc3:
🔥

PR #693, opened 2025-07-04
feat: add ssl_request_message macro for pattern matching

abc3:
This PR enforces strict and explicit pattern matching by replacing the raw binary pattern `<<_::64>>` with the semantically meaningful `ssl_request_message` macro, reducing the risk of accidental matches and improving readability

### DIFF
--- a/test/supavisor/protocol_test.exs
+++ b/test/supavisor/protocol_test.exs
@@ -122,6 +122,10 @@ defmodule Supavisor.ProtocolTest do
     assert @subject.ssl_request() == expected
   end
 
+  test "ssl_request_message/0" do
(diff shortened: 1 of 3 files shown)


test/supavisor/protocol_test.exs
```diff
@@ -122,6 +122,10 @@ defmodule Supavisor.ProtocolTest do
     assert @subject.ssl_request() == expected
   end
 
+  test "ssl_request_message/0" do
+    assert @subject.ssl_request_message == <<0, 0, 0, 8, 4, 210, 22, 47>>
```
v0idpwn: Just a nit:
```suggestion
    assert @subject.ssl_request_message() == <<0, 0, 0, 8, 4, 210, 22, 47>>
```

v0idpwn (approved):

PR #694, opened 2025-07-07
feat: improved named prepared statements support

v0idpwn:
**Disclaimer: this is a first draft, and the code is rough at places, there's missing functionality, etc**

### Concept
This PR implements the following architecture:
- ClientHandlers keep track of the prepared statements sent on them
- ClientHandlers update the prepared statement packets changing the statement names to "internal names" (this is necessary because clients may reuse the names in different connections)
- Prepared statement packets aren't sent directly to the db socket, but instead are sent to the DbHandler. Whenever the ClientHandler sends a `bind` packet to the DbHandler, it should also send the respective `parse` (which it has stored)
- DbHandler decides whether it needs to run "parse" or just "bind"
- DbHandler must intercept "parse" responses when the client didn't request them/it was made due to necessity

### Parsing/decoding
- To be able to handle the prepared statement packets, ClientHandlers need to decode every client packet (at least on transaction mode)
- To be able to intercept parse (and soon, close) responses, DbHandler needs to decode every server packet. 
- For both of them, we can avoid decoding the whole payload/decode only what's necessary (generally, we only need tagging, except for the prepared statement payloads from which we need more data).
- The decoding means more CPU usage, and it could also mean that we will need to optimize the decoding routines a bit further.

### Next steps
- Benchmarking/performance improvements
- Prepared statement limit client_handler side (hard limit)
- Prepared statement limit db_handler side (LRU)
- Make it work with read replicas too
- Improved error handling/ensure we conform to what client libraries (and Postgres) expect.
- Deterministic internal name -- right now I'm using `System.unique_integer`, but if we use pg_parser to give a name to the statement, we can reduce the number of unnecessary prepares.

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -26,6 +26,8 @@ defmodule Supavisor.ClientHandler do
     Tenants
   }
 
+  alias Supavisor.Protocol.PreparedStatements
+  alias Supavisor.Protocol.Client.Pkt, as: ClientPkt
   require Supavisor.Protocol.Server, as: Server
 
   @impl true
@@ -102,7 +104,8 @@ defmodule Supavisor.ClientHandler do
       active_count: 0,
       peer_ip: peer_ip,
       app_name: nil,
-      subscribe_retries: 0
+      subscribe_retries: 0,
+      prepared_statements: %{}
     }
 
     :gen_statem.enter_loop(__MODULE__, [hibernate_after: 5_000], :exchange, data)
@@ -420,6 +423,7 @@ defmodule Supavisor.ClientHandler do
            ),
          {:ok, opts} <- Supavisor.subscribe(sup, data.id) do
       manager_ref = Process.monitor(opts.workers.manager)
+
       data = Map.merge(data, opts.workers)
       db_pid = db_checkout(:both, :on_connect, data)
       data = %{data | manager: manager_ref, db_pid: db_pid, idle_timeout: opts.idle_timeout}
@@ -537,8 +541,7 @@ defmodule Supavisor.ClientHandler do
     {:keep_state_and_data, {:timeout, data.heartbeat_interval, :heartbeat_check}}
   end
 
-  def handle_event(kind, {proto, socket, msg}, state, data)
-      when proto in @proto and is_binary(msg) do
+  def handle_event(kind, {proto, socket, msg}, state, data) when proto in @proto do
     Logger.metadata(state: state)
 
     with {:next_state, next_state, new_data, actions} <- handle_data(kind, msg, state, data) do
@@ -548,6 +551,9 @@ defmodule Supavisor.ClientHandler do
         |> Enum.map(fn
           {:next_event, type, bin} when is_binary(bin) ->
             {:next_event, type, {proto, socket, bin}}
+
+          {:next_event, type, pkts} when is_list(pkts) ->
+            {:next_event, type, {proto, socket, pkts}}
 
           other ->
             other
@@ -1075,10 +1081,16 @@ defmodule Supavisor.ClientHandler do
   defp handle_data(:info, bin, :idle, %{pool: pid} = data) when is_pid(pid) do
     Logger.debug("ClientHandler: Receive query #{inspect(bin)}")
     db_pid = db_checkout(:both, :on_query, data)
-    handle_prepared_statements(db_pid, bin, data)
 
-    {:next_state, :busy, %{data | db_pid: db_pid, query_start: System.monotonic_time()},
-     {:next_event, :internal, bin}}
+    {:ok, bin_or_pkts, prepared_statements, _rest} = handle_prepared_statements(db_pid, bin, data)
+
+    {:next_state, :busy,
+     %{
+       data
+       | prepared_statements: prepared_statements,
+         db_pid: db_pid,
+         query_start: System.monotonic_time()
+     }, {:next_event, :internal, bin_or_pkts}}
   end
 
   defp handle_data(:info, bin, _, %{mode: :proxy} = data) do
@@ -1133,33 +1145,27 @@ defmodule Supavisor.ClientHandler do
     end
   end
 
-  @spec handle_prepared_statements({pid, pid, Supavisor.sock()}, binary, map) :: :ok | nil
-  defp handle_prepared_statements({_, pid, _}, bin, %{mode: :transaction} = data) do
-    with {:ok, payload} <- Client.get_payload(bin),
-         {:ok, statements} <- Supavisor.PgParser.statements(payload),
-         true <- statements in [["PrepareStmt"], ["DeallocateStmt"]] do
-      Logger.info("ClientHandler: Handle prepared statement #{inspect(payload)}")
+  @spec handle_prepared_statements({pid, pid, Supavisor.sock()}, binary, map) ::
+          {:ok, [ClientPkt.t()] | binary, map, binary}
+  defp handle_prepared_statements(
+         {_, _pid, _},
+         bin,
+         %{mode: :transaction} = data
+       ) do
+    stmts = data.prepared_statements
 
-      GenServer.call(data.pool, :get_all_workers)
-      |> Enum.each(fn
-        {_, ^pid, _, [Supavisor.DbHandler]} ->
-          Logger.debug("ClientHandler: Linked DbHandler #{inspect(pid)}")
-          nil
+    {:ok, pkts, rest} = Client.decode(bin)
 
-        {_, pool_proc, _, [Supavisor.DbHandler]} ->
-          Logger.debug(
-            "ClientHandler: Sending prepared statement change #{inspect(payload)} to #{inspect(pool_proc)}"
-          )
-
-          send(pool_proc, {:handle_ps, payload, bin})
+    {stmts, pkts} =
+      Enum.reduce(pkts, {stmts, []}, fn pkt, {stmts, pkts} ->
+        {stmts, pkt} = PreparedStatements.handle_pkt(stmts, pkt)
+        {stmts, [pkt | pkts]}
       end)
-    else
-      error ->
-        Logger.debug("ClientHandler: Skip prepared statement #{inspect(error)}")
-    end
+
+    {:ok, Enum.reverse(pkts), stmts, rest}
   end
 
-  defp handle_prepared_statements(_, _, _), do: nil
+  defp handle_prepared_statements(_, bin, data), do: {:ok, bin, data.prepared_statements, ""}
 
   @spec handle_actions(map) :: [{:timeout, non_neg_integer, atom}]
   defp handle_actions(%{} = data) do
@@ -1195,8 +1201,8 @@ defmodule Supavisor.ClientHandler do
 
   def maybe_change_log(_), do: :ok
 
-  @spec sock_send_maybe_active_once(binary(), map()) :: :ok | {:error, term()}
-  def sock_send_maybe_active_once(bin, data) do
+  @spec sock_send_maybe_active_once(binary() | [Pkt.t()], map()) :: :ok | {:error, term()}
+  def sock_send_maybe_active_once(bin, data) when is_binary(bin) do
     Logger.debug("ClientHandler: Send maybe active once")
     active_count = data.active_count
 
@@ -1206,6 +1212,36 @@ defmodule Supavisor.ClientHandler do
     end
 
     HandlerHelpers.sock_send(elem(data.db_pid, 2), bin)
+  end
+
+  # TODO: we can probably optimize this if we:
+  # - chunk all non-prepared statements messages in an iolist before sending it
+  # - potentially chunk prepared statements messages too
+  # e.g.:
+  # [msg1, msg2, msg3, parse, bind, sync, msg4, ms5] could be:
+  # sock:send([msg1, msg2, msg3])
+  # db_handler:call([parse, bind, sync])
+  # sock:send([msg4, msg5])
+  # Since we are using nodelay: true, this probably should give us a bit more
+  # thoughtput at a small price in processing
+  def sock_send_maybe_active_once(pkts, data) do
+    {_pool, db_handler, db_sock} = data.db_pid
+    active_count = data.active_count
+
+    if active_count > @switch_active_count do
+      Logger.debug("ClientHandler: Activate socket #{inspect(active_count)}")
+      HandlerHelpers.active_once(data.sock)
+    end
+
+    Enum.each(pkts, fn pkt ->
+      case pkt do
+        %ClientPkt{tag: tag} when tag in [:close_message, :bind_message, :parse_message] ->
+          Supavisor.DbHandler.handle_prepared_statement_pkt(db_handler, pkt)
+
+        %ClientPkt{bin: bin} ->
+          HandlerHelpers.sock_send(db_sock, bin)
+      end
+    end)
   end
 
   @spec timeout_subscribe_or_terminate(map()) :: :gen_statem.handle_event_result()
--- a/lib/supavisor/db_handler.ex
+++ b/lib/supavisor/db_handler.ex
@@ -8,6 +8,9 @@ defmodule Supavisor.DbHandler do
 
   @behaviour :gen_statem
 
+  alias Supavisor.Protocol.Client.Pkt, as: ClientPkt
+  alias Supavisor.Protocol.Server.Pkt, as: ServerPkt
+
   alias Supavisor.{
     ClientHandler,
     HandlerHelpers,
@@ -33,6 +36,11 @@ defmodule Supavisor.DbHandler do
 
   @spec checkin(pid()) :: :ok
   def checkin(pid), do: :gen_statem.cast(pid, :checkin)
+
+  @spec handle_prepared_statement_pkt(pid, term()) :: :ok
+  def handle_prepared_statement_pkt(pid, pkt) do
+    :gen_statem.call(pid, {:handle_prepared_statement, pkt})
+  end
 
   @spec get_state_and_mode(pid()) :: {:ok, {state, Supavisor.mode()}} | {:error, term()}
   def get_state_and_mode(pid) do
@@ -73,6 +81,9 @@ defmodule Supavisor.DbHandler do
         server_proof: nil,
         stats: %{},
         client_stats: %{},
+        prepared_statements: MapSet.new(),
+        pending_bin: "",
+        intercept_parse_resps: 0,
         mode: args.mode,
         replica_type: args.replica_type,
         reply: nil,
@@ -158,7 +169,7 @@ defmodule Supavisor.DbHandler do
   end
 
   def handle_event(:info, {proto, _, bin}, :authentication, data) when proto in @proto do
-    dec_pkt = Server.decode(bin)
+    {:ok, dec_pkt, _} = Server.decode(bin)
     Logger.debug("DbHandler: dec_pkt, #{inspect(dec_pkt, pretty: true)}")
 
     resp = Enum.reduce(dec_pkt, %{}, &handle_auth_pkts(&1, &2, data))
@@ -224,7 +235,7 @@ defmodule Supavisor.DbHandler do
     if buff != [] do
       Logger.debug("DbHandler: Buffer is not empty, try to send #{IO.iodata_length(buff)} bytes")
       buff = Enum.reverse(buff)
-      :ok = sock_send(data.sock, buff)
+      :ok = HandlerHelpers.sock_send(data.sock, buff)
     end
 
     {:next_state, :busy, %{data | buffer: []}}
@@ -240,7 +251,7 @@ defmodule Supavisor.DbHandler do
       )
 
       buff = Enum.reverse(buff)
-      :ok = sock_send(data.sock, buff)
+      :ok = HandlerHelpers.sock_send(data.sock, buff)
     end
 
     {:keep_state, %{data | anon_buffer: []}}
@@ -260,7 +271,8 @@ defmodule Supavisor.DbHandler do
   def handle_event(:info, {proto, _, bin}, _, %{replica_type: :read} = data)
       when proto in @proto do
     Logger.debug("DbHandler: Got read replica message #{inspect(bin)}")
-    pkts = Server.decode(bin)
+    {:ok, pkts, rest} = Server.decode(<<data.pending_bin::binary, bin::binary>>)
+    data = %{data | pending_bin: rest}
 
     resp =
       cond do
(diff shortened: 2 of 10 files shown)


abc3:
> To be able to handle the prepared statement packets, ClientHandlers need to decode every client packet (at least on transaction mode)

Pattern matching on messages starting with `?P` and `?B` can isolate prepared statement handling from cases where the client doesn't use them.

lib/supavisor/db_handler.ex
```diff
@@ -34,6 +37,11 @@ defmodule Supavisor.DbHandler do
   @spec checkin(pid()) :: :ok
   def checkin(pid), do: :gen_statem.cast(pid, :checkin)
 
+  @spec handle_prepared_statement_pkts(pid, [PreparedStatements.handled_pkt()]) :: :ok
+  def handle_prepared_statement_pkts(pid, pkts) do
+    :gen_statem.call(pid, {:handle_ps_pkts, pkts})
```
abc3: by default, the timeout is set to `:infinity`, which can lead to deadlocks

```suggestion
    :gen_statem.call(pid, {:handle_ps_pkts, pkts}, 15_000)
```

lib/supavisor/protocol.ex
```diff
@@ -0,0 +1,20 @@
...
+
+  defp do_split_pkts(
+         <<_char::8, pkt_len::32, _rest::binary-size(pkt_len - 4), rest2::binary>> = bin,
+         acc
+       ) do
+    do_split_pkts(rest2, [:binary.part(bin, 0, pkt_len + 1) | acc])
+  end
+
+  defp do_split_pkts(rest, acc) do
+    {:lists.reverse(acc), rest}
```
abc3, line 22:
```suggestion
    {Enum.reverse(acc), rest}
```

lib/supavisor/protocol/prepared_statements.ex
```diff
@@ -0,0 +1,179 @@
...
+  @type pkt() :: binary()
+
+  @type handled_pkt() ::
+          {:parse_pkt, statement_name(), pkt()}
+          | {:bind_pkt, statement_name(), bind_pkt :: pkt(), parse_pkt :: pkt()}
+          | {:close_pkt, statement_name(), pkt()}
+          | pkt()
+
+  @client_limit 100
+  @backend_limit 200
```
abc3: probably worth configuring it per tenant or globally via envs

lib/supavisor/protocol/prepared_statements.ex
```diff
@@ -0,0 +1,179 @@
...
+        handle_describe_message(client_statements, len, rest)
+
+      # Query message (Q)
+      <<?Q, len::32, rest::binary>> ->
+        handle_simple_query_message(client_statements, binary, len, rest)
+
+      # All other messages pass through unchanged
+      _ ->
+        {:ok, client_statements, binary}
+    end
```
abc3: I would handle that in `Protocol.do_split_pkts`

lib/supavisor/client_handler.ex
```diff
@@ -554,6 +558,9 @@ defmodule Supavisor.ClientHandler do
           {:next_event, type, bin} when is_binary(bin) ->
             {:next_event, type, {proto, socket, bin}}
 
+          {:next_event, type, pkts} when is_list(pkts) ->
+            {:next_event, type, {proto, socket, pkts}}
+
```
abc3:
```suggestion
          {:next_event, type, bin_or_pkts} when is_binary(bin_or_pkts) or is_list(bin_or_pkts) ->
            {:next_event, type, {proto, socket, bin_or_pkts}}
```

lib/supavisor/db_handler.ex
```diff
@@ -694,4 +724,98 @@ defmodule Supavisor.DbHandler do
...
+         expected_parse_resps: data.expected_parse_resps + 1
+     }}
+  end
+
+  defp remove_exceeding(prepared_statements) do
+    limit = PreparedStatements.backend_limit()
+
+    if MapSet.size(prepared_statements) >= limit do
+      count = div(limit, 5)
+      to_remove = Enum.take_random(prepared_statements, count) |> MapSet.new()
```
abc3, line 778: maybe extend the item with the latest timestamp and remove using LRU

lib/supavisor/client_handler.ex
```diff
@@ -1119,52 +1127,89 @@ defmodule Supavisor.ClientHandler do
...
-    Logger.debug("ClientHandler: Forward query to db #{inspect(bin)} #{inspect(data.db_pid)}")
+  defp handle_data(_, data_to_send, :busy, data) do
+    Logger.debug(
+      "ClientHandler: Forward query to db #{inspect(data_to_send)} #{inspect(data.db_pid)}"
+    )
 
-    case sock_send_maybe_active_once(bin, data) do
-      :ok ->
-        {:keep_state, %{data | active_count: data.active_count + 1}}
+    case handle_client_pkts(<<data.pending::binary, data_to_send::binary>>, data) do
```
abc3: iolist_to_binary would probably be faster (if pending also iolist)

lib/supavisor/client_handler.ex
```diff
@@ -1119,52 +1127,89 @@ defmodule Supavisor.ClientHandler do
...
           data.sock,
-          Server.error_message("XX000", "Error while sending query")
+          Server.error_message("XX000", message_text)
         )
 
-        {:stop, {:shutdown, :send_query_error}}
+        {:stop, {:shutdown, :max_prepared_statements}}
+
+      {:error, :prepared_statement_on_simple_query} ->
+        message_text = "Supavisor only supports PREPARE/EXECUTE using the Extended Query Protocol"
```
abc3: PREPARE/EXECUTE is part of the simple protocol

```suggestion
      {:error, :prepared_statement_on_simple_query} ->
        message_text = "Supavisor only supports prepared statements created via the Extended Query Protocol"
```

lib/supavisor/db_handler.ex
```diff
@@ -694,4 +724,98 @@ defmodule Supavisor.DbHandler do
...
+  # If we received a bind without a parse, we need to intercept the parse response, otherwise,
+  # the client will receive an unexpected message.
+  defp handle_prepared_statement_pkt({:bind_pkt, stmt_name, pkt, parse_pkt}, {iodata, data}) do
+    new_ps? = stmt_name not in data.prepared_statements
+
+    if new_ps? and parse_pkt do
+      new_data = %{
+        data
+        | intercept_parse_resps: data.intercept_parse_resps + 1,
+          prepared_statements: MapSet.put(data.prepared_statements, stmt_name)
```
abc3, line 723: since different clients_handlers can use the same db_handler, the key should probably be a pair of the caller pid and the prepared statement name

lib/supavisor/client_handler.ex
```diff
@@ -1213,6 +1266,35 @@ defmodule Supavisor.ClientHandler do
...
+      Logger.debug("ClientHandler: Activate socket #{inspect(active_count)}")
+      HandlerHelpers.active_once(data.sock)
+    end
+
+    pkts
+    |> Enum.chunk_by(&is_tuple/1)
+    |> Enum.reduce_while(:ok, fn chunk, _acc ->
+      case chunk do
+        [t | _] = prepared_pkts when is_tuple(t) ->
+          Supavisor.DbHandler.handle_prepared_statement_pkts(db_handler, prepared_pkts)
```
abc3, line 1364: `:gen_statem.call/2` is called twice per PS query: once for `bind` and once for `close`, yeah?

lib/supavisor/protocol/prepared_statements.ex
```diff
@@ -0,0 +1,179 @@
...
+
+          new_client_statements = Map.put(client_statements, client_side_name, prepared_statement)
+          {:ok, new_client_statements, {:parse_pkt, server_side_name, new_bin}}
+        end
+    end
+  end
+
+  defp handle_bind_message(client_statements, len, rest) do
+    {_portal_name, after_portal} = extract_null_terminated_string(rest)
+    {client_side_name, packet_after_client_name} = extract_null_terminated_string(after_portal)
```
abc3: likely need to skip further processing if the PS is unnamed

lib/supavisor/protocol/prepared_statements.ex
```diff
@@ -0,0 +1,179 @@
...
+
+    case Map.get(client_statements, client_side_name) do
+      %PreparedStatement{name: server_side_name, parse_pkt: parse_pkt} ->
+        new_len = len + (byte_size(server_side_name) - byte_size(client_side_name))
+
+        new_bin =
+          <<?B, new_len::32, 0, server_side_name::binary, 0, packet_after_client_name::binary>>
```
abc3: Use iodata in such places? It seems that this bin will just be sent over the socket

lib/supavisor/protocol/prepared_statements.ex
```diff
@@ -0,0 +1,179 @@
...
+
+        {:ok, client_statements, {:bind_pkt, server_side_name, new_bin, parse_pkt}}
+
+      nil ->
+        # Unknown statement name - use empty string
+        # This probably should be an error. Need to double check it.
+        new_len = len + (0 - byte_size(client_side_name))
```
abc3:
```suggestion
        new_len = len - byte_size(client_side_name)
```

lib/supavisor/protocol/client.ex
```diff
@@ -163,6 +171,10 @@ defmodule Supavisor.Protocol.Client do
     :undef
   end
 
+  def discard_all_message do
+    <<?Q, 15, "DISCARD ALL">>
```
abc3: precompiled may be slightly more performant

lib/supavisor/db_handler.ex
```diff
@@ -325,16 +348,28 @@ defmodule Supavisor.DbHandler do
...
     end
   end
 
-  def handle_event(:info, {:handle_ps, payload, bin}, _state, data) do
-    Logger.notice("DbHandler: Apply prepare statement change #{inspect(payload)}")
+  def handle_event({:call, from}, {:handle_ps_pkts, pkts}, _state, data) do
+    {iodata, data} =
+      Enum.reduce(pkts, {[], data}, fn pkt, {iodata, data} ->
+        handle_prepared_statement_pkt(pkt, {iodata, data})
+      end)
```
abc3:
```suggestion
      Enum.reduce(pkts, {[], data}, fn pkt, acc ->
        handle_prepared_statement_pkt(pkt, acc)
      end)
```

lib/supavisor/protocol/server.ex
```diff
@@ -179,13 +210,21 @@ defmodule Supavisor.Protocol.Server do
 
   # Internal functions
 
-  @spec decode(binary(), list()) :: [Pkt.t()]
+  @spec decode(binary(), list()) :: {:ok, [Pkt.t()], rest :: binary()}
```
abc3, line 221: nitpick: no need `:ok` if the function cannot return `:error`

lib/supavisor/protocol/server.ex
```diff
@@ -38,24 +44,49 @@ defmodule Supavisor.Protocol.Server do
...
 
       _ ->
-        {:error, :bad_packet}
+        {:error, :incomplete}
     end
   end
 
+  def decode_pkt(<<_char::integer-8, _pkt_len::integer-32, _rest>>) do
+    {:error, :incomplete}
+  end
```
abc3: it seems that the clause above matches the same, so the current one will never be called

lib/supavisor/protocol/prepared_statements.ex
v0idpwn: I blame claude here

lib/supavisor/protocol/client.ex
v0idpwn: Oh, this is remains from me trying other approach, ended up not even using this one.

lib/supavisor/protocol/server.ex
v0idpwn: Also remaisn from previous approach, good catch

lib/supavisor/db_handler.ex
v0idpwn: Yup! This is something that I'm considering as an improvement, but probably won't do in this PR

lib/supavisor/db_handler.ex
v0idpwn: Not now because the name is guaranteed unique, but if we start going for fixed names, then yes, we need

lib/supavisor/client_handler.ex
v0idpwn: No! I'm calling it once for a "chunk" of PS, that's why we do the `chunk_by` before

So if we get `close, parse, bind` (postgrex loves to send these) all of them are handled in a single call

lib/supavisor/protocol.ex
v0idpwn: I was thinking if I should go straight for `:lists` instead of `Enum` in these very hot paths 🤔, ended up doing 50/50, need to decide if should use `:lists` or `Enum` everywhere

lib/supavisor/protocol/prepared_statements.ex
v0idpwn: I think this is a good call but will probably hold it for another PR

lib/supavisor/protocol/prepared_statements.ex
v0idpwn: The good part is one less pass, but it also makes stuff harder to reason about (and will require refactoring since DbHandler also uses `split_pkts`, but for backend messages). We can check if this pass makes significant difference, if we can scrap a few microseconds doing it, then we cut it. But I suggest doing it after in an improvement PR, then we can consider more carefully if the performance gain is worth the code impact.

lib/supavisor/client_handler.ex
v0idpwn: 🤦🏻‍♂️

lib/supavisor/client_handler.ex
v0idpwn: `data.pending` is always a binary, I did a little [bench](https://gist.github.com/v0idpwn/fad46c246efa0bd1e57d581875a7d6bc) on it and it seems <<a::binary, b::binary>> is faster in this case.

lib/supavisor/protocol/prepared_statements.ex
v0idpwn: !!!

v0idpwn: This is actually huge! Not handling it here is what caused extra latency when using extended query protocol **without** named prepared statements.

v0idpwn:
Just duplicating this info since it's in the middle of a lot of threads: after some improvements, it seems like there's no (significant) latency impact when running pgbench `-M extended`.

lib/supavisor/protocol/prepared_statements.ex
v0idpwn: I added it in the next steps in PR description and will resolve this comment (PRs welcome on this btw 👀)

lib/supavisor/db_handler.ex
v0idpwn: I added it in the next steps in PR description and will resolve this comment (PRs welcome on this btw 👀)

lib/supavisor/db_handler.ex
```diff
@@ -307,10 +327,13 @@ defmodule Supavisor.DbHandler do
       # after which it will unlink the direct db connection process from itself.
       data =
         if data.mode == :transaction do
-          ClientHandler.db_status(data.caller, :ready_for_query, bin)
+          # for some reason, if we send the data **before** doing the `ready_for_query` cast,
+          # we get race condition that sometimes causes msgs to be sent to the wrong socket
+          ClientHandler.db_status(data.caller, :ready_for_query)
+          data = intercept_or_send_pkts(pkts, data)
```
v0idpwn: I don't get why this happens. For me it seems like we want to send stuff and only then call `db_status`. But if I do it it breaks things.

lib/supavisor/protocol.ex
abc3: Enum looks more consistent for Elixir

lib/supavisor/client_handler.ex
abc3: hmmm, i thought i saw a function call for each message 🤷

abc3:
jfyi pgbench perf needs to be checked with MIX_ENV=prod, because in production, debug logging is purged

lib/supavisor/db_handler.ex
abc3: 🤔🤔🤔

lib/supavisor/protocol/prepared_statements.ex
abc3: If handled in do_split_pkts, you can potentially skip PS processing when it's not needed

lib/supavisor/db_handler.ex
```diff
@@ -723,37 +722,59 @@ defmodule Supavisor.DbHandler do
...
-        case {pkt, data} do
-          {Server.parse_complete_message(), %{expected_parse_resps: n}} when n > 0 ->
-            {[pkt | acc_bins], %{data | expected_parse_resps: n - 1}}
+  @spec handle_server_messages([binary()], map()) :: map()
+  defp handle_server_messages(pkts, data) do
+    {packets_to_send, updated_data} = process_messages(pkts, [], data)
 
-          {Server.parse_complete_message(), %{intercept_parse_resps: n}} when n > 0 ->
-            {acc_bins, %{data | intercept_parse_resps: n - 1}}
+    unless packets_to_send == [] do
```
abc3: [unless](https://github.com/elixir-lang/elixir/blob/7b20c281d521aa7aa2ad2baa1e9ae6c579d79d0c/lib/elixir/lib/kernel.ex#L3963-L3965) will be depreciated

lib/supavisor/db_handler.ex
```diff
@@ -723,37 +722,59 @@ defmodule Supavisor.DbHandler do
...
+      HandlerHelpers.sock_send(data.client_sock, Enum.reverse(packets_to_send))
```
abc3: 🔥

lib/supavisor/db_handler.ex
```diff
@@ -396,20 +396,16 @@ defmodule Supavisor.DbHandler do
...
 
-    if state == :busy or data.mode == :session do
-      HandlerHelpers.sock_send(data.sock, Server.terminate_message())
-      :gen_tcp.close(elem(data.sock, 1))
-      {:stop, {:client_handler_down, data.mode}}
-    else
-      {:keep_state, %{data | caller: nil, buffer: []}}
-    end
+    HandlerHelpers.sock_send(data.sock, Server.terminate_message())
+    :gen_tcp.close(elem(data.sock, 1))
```
abc3:
```suggestion
    HandlerHelpers.sock_close(data.sock)
```

lib/supavisor/protocol/prepared_statements.ex
```diff
@@ -19,6 +20,7 @@ defmodule Supavisor.Protocol.PreparedStatements do
 
   @client_limit 100
   @backend_limit 200
+  @client_memory_limit 1_000_000
```
abc3: would be nice for clarity
```suggestion
  @client_memory_limit_bytes 1_000_000
```

.tool-versions
```diff
@@ -1,4 +1,4 @@
 rust 1.84.1
 elixir 1.18.2-otp-27
 nodejs 19.9.0
-erlang 27.2
+erlang 27.2
```
abc3:
```suggestion
erlang 27.2

```

lib/supavisor/db_handler.ex
```diff
@@ -700,6 +730,104 @@ defmodule Supavisor.DbHandler do
   def reconnect_timeout(_),
     do: @reconnect_timeout
 
+  @spec handle_server_messages(binary(), map()) :: map()
+  defp handle_server_messages(bin, data) do
+    if Supavisor.FeatureFlag.enabled?(data.tenant_feature_flags, "named_prepared_statements") do
```
abc3:
```suggestion
    if FeatureFlag.enabled?(data.tenant_feature_flags, "named_prepared_statements") do
```

lib/supavisor/helpers.ex
```diff
@@ -399,4 +395,27 @@ defmodule Supavisor.Helpers do
...
+        true
+
+      "false" ->
+        false
+
+      "1" ->
+        true
+
+      "0" ->
+        false
```
abc3: a bit shorter
```suggestion
      value when value in ["true", "1"] ->
        true

      value when value in ["false", "0"] ->
        false
```

native/pgparser/src/lib.rs
```diff
@@ -10,4 +10,11 @@ fn statement_types(query: &str) -> Result<Vec<String>, String> {
     Ok(message)
 }
 
+#[rustler::nif]
+fn fingerprint(query: &str) -> Result<String, String> {
+    pg_query::fingerprint(query)
+        .map(|fingerprint| Ok(fingerprint.hex))
+        .map_err(|_| "Error fingerprinting query")?
+}
+
```
abc3: probably don't need anymore

abc3 (approved):
🚀 🚀 🚀

lib/supavisor/protocol/prepared_statements.ex
```diff
@@ -0,0 +1,227 @@
...
+      {"", _} ->
+        {:ok, client_statements, <<?P, len::32, payload::binary>>}
+
+      {client_side_name, remaining} ->
+        cond do
+          Storage.statement_count(client_statements) >= @client_limit ->
+            {:error, :max_prepared_statements}
+
+          Storage.statement_memory(client_statements) > @client_memory_limit_bytes ->
+            {:error, :max_prepared_statements_memory}
```
abc3: does here `client_statements` include a new statement?

lib/supavisor/db_handler.ex
```diff
@@ -4,15 +4,22 @@ defmodule Supavisor.DbHandler do
...
+
   require Logger
+  require Supavisor.Protocol.Server, as: Server
+  require Supavisor.Protocol.MessageStreamer, as: MessageStreamer
 
-  @behaviour :gen_statem
+  alias Supavisor.Protocol.PreparedStatements
 
   alias Supavisor.{
     ClientHandler,
```
v0idpwn, line 16:
```suggestion
    ClientHandler,
    FeatureFlag,
```

lib/supavisor/protocol/prepared_statements.ex
v0idpwn: Great point! Will change to this pkt size + `client_statement` size.

v0idpwn: Done in https://github.com/supabase/supavisor/pull/694/commits/6a752d477eb897e9d77e37875abfb700ad6dce97

PR #701, opened 2025-07-17
feat: refresh PromEx application metrics on soft deploy

v0idpwn:
The PromEx application metrics are "manual", so they are only emitted on application start or upon calling
`PromEx.ManualMetricsManager.refresh_metrics/1`. Since we don't do a restart on soft deploys, the relevant metrics (most importantly, application version) stay outdated in Grafana until a restart happens. This commit adds a refresh in the update script.

### DIFF
--- a/lib/supavisor/hot_upgrade.ex
+++ b/lib/supavisor/hot_upgrade.ex
@@ -154,7 +154,10 @@ defmodule Supavisor.HotUpgrade do
   end
 
   def reinit_manual_metrics do
-    PromEx.ManualMetricsManager.refresh_metrics(PromEx.Plugins.Application)
+    spawn(fn ->
+      :timer.sleep(60_000)
+      PromEx.ManualMetricsManager.refresh_metrics(PromEx.Plugins.Application)
+    end)
   end
 
   @spec enc(term) :: fun
--- a/lib/supavisor/hot_upgrade.ex
+++ b/lib/supavisor/hot_upgrade.ex
@@ -94,6 +94,7 @@ defmodule Supavisor.HotUpgrade do
   def reint_funs do
     reinit_pool_args()
     reinit_auth_query()
+    reinit_manual_metrics()
   end
 
   def reinit_pool_args do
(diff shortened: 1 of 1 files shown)


lib/supavisor/hot_upgrade.ex
```diff
@@ -152,6 +153,10 @@ defmodule Supavisor.HotUpgrade do
     end)
   end
 
+  def reinit_manual_metrics do
+    PromEx.ManualMetricsManager.refresh_metrics(PromEx.Plugins.Application)
```
v0idpwn:
```suggestion
    spawn(fn ->
      :timer.sleep(60_000)
      PromEx.ManualMetricsManager.refresh_metrics(PromEx.Plugins.Application)
    end)
```

lib/supavisor/hot_upgrade.ex
```diff
@@ -152,6 +153,13 @@ defmodule Supavisor.HotUpgrade do
     end)
   end
 
+  def reinit_manual_metrics do
+    spawn(fn ->
+      :timer.sleep(60_000)
+      PromEx.ManualMetricsManager.refresh_metrics(PromEx.Plugins.Application)
+    end)
+  end
```
v0idpwn, line 161: Unfortunately, I didn't find any api that ensures that it runs **after** the relup was completed.

abc3: how about restarting prom_ex?

```elixir
[Application.stop(:prom_ex), Application.start(:prom_ex)]
```

v0idpwn: I think running `PromEx.ManualMetricsManager.refresh_metrics/1` is the correct thing to do, what I don't like is how I'm doing (timer based), I wish I could do it in a hook when the update completes successfully.

abc3: You can extend Updater to invoke some new function from Supavisor after it upgrades the release

v0idpwn: I guess this will be the way. Didn't want to touch updater but I dislike the timer approach too much)

PR #703, opened 2025-07-17
feat: support manually written appups

v0idpwn:
This change allows us to write (and version) appups instead of using distillery to gen Supavisor's appups.

It's something that we may need in some cases, e.g.: the update from v2.0.21 to v2.1.x requires a few dependency bumps. If we write the appups for these dependencies, we can run a soft deploy.

### Expected flow

**Whenever you can, use the auto-generated appup**, with `UPGRADE_FROM=x.y.z mix release`. When you can't, you may use manual appups:

1. Determine the supavisor version upgrade path. E.g.: `2.0.21-2.1.3`. Create a directory for this upgrade path on `relups`. Inside it, create an `appups` directory.
2. Generate the necessary appups with `mix supavisor.gen.appup`. For example, `mix supavisor.gen.appup --app=phoenix_live_view --from=0.20.17 --to=1.0.2 --path=relups/2.0.21-2.1.3/appups`. At this point, the directory structure will look somewhat like this:
```
relups
└── 2.0.21-2.1.3
    └── appups
        ├── libcluster-3.5.0.appup
        ├── meck-1.0.0.appup
        ├── phoenix_live_view-1.0.2.appup
        ├── supavisor-2.1.3.appup
        └── telemetry_metrics-1.1.0.appup
```
3. Run `UPGRADE_FROM=x.y.z mix release` as usual. The assembled release will use the appups from the directory above. If you missed some appup, it will complain here.
4. **Review/test the upgrade process extensively**. A broken relup can cause issues, and since the rollback is part of the relup, chances are that if the upgrade is broken, the downgrade is broken as well.
5. Commit the appups and soft deploy.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -134,7 +134,7 @@ defmodule Supavisor.MixProject do
       rel_content = File.read!(Path.join(release.version_path, "supavisor.rel"))
       :ok = File.write!(path, rel_content)
 
-      # If appups directory exists manually, don't run the tasks, and instead
+      # If appups directory exist, don't run the task to generate, and instead
       # use the existing appups.
       appups_path = Path.join(["relups", "#{from}-#{vsn}", "appups"])
 
--- a/mix.exs
+++ b/mix.exs
@@ -126,15 +126,47 @@ defmodule Supavisor.MixProject do
   end
 
   defp upgrade(release) do
-    from = System.get_env("UPGRADE_FROM")
+    from = System.get_env("UPGRADE_FROM", "")
 
-    if from && from != "" do
+    if from != "" do
       vsn = release.version
       path = Path.join([release.path, "releases", "supavisor-#{vsn}.rel"])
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -126,15 +126,47 @@ defmodule Supavisor.MixProject do
...
+    if from != "" do
       vsn = release.version
       path = Path.join([release.path, "releases", "supavisor-#{vsn}.rel"])
       rel_content = File.read!(Path.join(release.version_path, "supavisor.rel"))
-
-      Mix.Task.run("supavisor.gen.appup", ["--from=" <> from, "--to=" <> vsn])
       :ok = File.write!(path, rel_content)
+
+      # If appups directory exists manually, don't run the tasks, and instead
+      # use the existing appups.
```
v0idpwn:
```suggestion
      # If appups directory exist, don't run the task to generate, and instead
      # use the existing appups.
```

PR #706, opened 2025-07-20
perf: constant number of ranch instances

v0idpwn:
Instead of starting one ranch instance per pool, use the same two ranch instances for all pools.

Each ranch was starting a minimum of 10 acceptors (and consequently, 10 connection supervisors). With 10_000 pools, these are 200_000 processes, which consume a sizeable amount of memory and resources. They also added complexity in managing the separate ranch instances (need to start and finish them at appropriate times, specially because they weren't linked to the pool).

The ranch acceptors/connection supervisors aren't a bottleneck, and if they were, we can control the number of acceptors/supervisors through configuration. In synthesis, there's no benefit in starting bringing multiple ranch instances, only the resource consumption drawback.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -159,6 +159,7 @@ if config_env() != :test do
     proxy_port_transaction:
       System.get_env("PROXY_PORT_TRANSACTION", "6543") |> String.to_integer(),
     proxy_port_session: System.get_env("PROXY_PORT_SESSION", "5432") |> String.to_integer(),
+    proxy_port_internal: System.get_env("PROXY_PORT_INTERNAL", "5413") |> String.to_integer(),
     proxy_port: System.get_env("PROXY_PORT", "5412") |> String.to_integer(),
     prom_poll_rate: System.get_env("PROM_POLL_RATE", "15000") |> String.to_integer(),
     global_upstream_ca: upstream_ca,
--- a/lib/supavisor/application.ex
+++ b/lib/supavisor/application.ex
@@ -48,15 +48,21 @@ defmodule Supavisor.Application do
         {Supavisor.SignalHandler, []}
       )
 
-    proxy_ports = [
-      {:pg_proxy_transaction, Application.get_env(:supavisor, :proxy_port_transaction),
-       :transaction, Supavisor.ClientHandler},
-      {:pg_proxy_session, Application.get_env(:supavisor, :proxy_port_session), :session,
-       Supavisor.ClientHandler},
-      {:pg_proxy, Application.get_env(:supavisor, :proxy_port), :proxy, Supavisor.ClientHandler}
-    ]
+    proxy_ports =
+      [
+        {:pg_proxy_transaction, Application.get_env(:supavisor, :proxy_port_transaction),
+         %{mode: :transaction, local: false}, Supavisor.ClientHandler},
+        {:pg_proxy_session, Application.get_env(:supavisor, :proxy_port_session),
+         %{mode: :session, local: false}, Supavisor.ClientHandler},
+        {{:pg_proxy_internal, :session}, 0, %{mode: :session, local: true},
+         Supavisor.ClientHandler},
+        {{:pg_proxy_internal, :transaction}, 0, %{mode: :transaction, local: true},
+         Supavisor.ClientHandler},
+        {:pg_proxy, Application.get_env(:supavisor, :proxy_port), %{mode: :proxy, local: false},
+         Supavisor.ClientHandler}
+      ]
 
-    for {key, port, mode, handler} <- proxy_ports do
+    for {key, port, opts, handler} <- proxy_ports do
       case :ranch.start_listener(
              key,
              :ranch_tcp,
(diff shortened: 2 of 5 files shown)


config/runtime.exs
```diff
@@ -159,6 +159,7 @@ if config_env() != :test do
     proxy_port_transaction:
       System.get_env("PROXY_PORT_TRANSACTION", "6543") |> String.to_integer(),
     proxy_port_session: System.get_env("PROXY_PORT_SESSION", "5432") |> String.to_integer(),
+    proxy_port_internal: System.get_env("PROXY_PORT_INTERNAL", "5413") |> String.to_integer(),
```
v0idpwn:
```suggestion
```

filipecabaco (approved):

abc3:
Yeah, a good initiative. Just adding some context for why it was originally done this way.

The initial approach aimed to provide tenant-level network isolation and reduce the risk of cross-tenant access through internal proxying. It also allowed us to observe TCP proxying behavior under these conditions.

We considered both unique ports per tenant and shared ones. For the shared setup, the idea was to maintain a pool of Ranch instances and select one via something like `:erlang.phash2(tenant_id, ...)` to avoid a single point of failure. We started with unique ports and planned to revisit these options later.

Creating Ranch instances is resource-intensive mainly during initialization. When idle, they add negligible overhead.

Here are load charts with 10k local Ranch instances. Scheduler usage is about the same as without them, and memory stays around 1GB

Screenshot 2025-07-21 at 10 26 09

config/runtime.exs
```diff
@@ -158,6 +158,7 @@ reconnect_retries =
 
 if config_env() != :test do
   config :supavisor,
+    local_server_shards: System.get_env("LOCAL_SERVER_SHARDS", "4") |> String.to_integer(),
```
abc3: nitpick: "proxy" is clearer
```suggestion
    local_proxy_shards: System.get_env("LOCAL_PROXY_SHARDS", "4") |> String.to_integer(),
```

lib/supavisor/application.ex
```diff
@@ -48,15 +48,29 @@ defmodule Supavisor.Application do
...
-      {:pg_proxy, Application.get_env(:supavisor, :proxy_port), :proxy, Supavisor.ClientHandler}
-    ]
+    local_server_shards = Application.fetch_env!(:supavisor, :local_server_shards)
+
+    session_shards = for shard <- 0..(local_server_shards - 1) do
+      {{:pg_proxy_internal, :session, shard}, 0, %{mode: :session, local: true, shard: shard},
+       Supavisor.ClientHandler}
+    end
+
+    transaction_shards = for shard <- 0..(local_server_shards - 1) do
```
abc3:
```suggestion

    shards = 0..(local_server_shards - 1)

    session_shards = for shard <- shards do
      {{:pg_proxy_internal, :session, shard}, 0, %{mode: :session, local: true, shard: shard},
       Supavisor.ClientHandler}
    end

    transaction_shards = for shard <- shards do
```

v0idpwn: Did it slightly differently on https://github.com/supabase/supavisor/pull/706/commits/c962967f7dfb49e3c6009a3d97c7c8e1421b1124

lib/supavisor/application.ex
```diff
@@ -66,10 +76,10 @@ defmodule Supavisor.Application do
                socket_opts: [port: port, keepalive: true]
              },
              handler,
-             %{mode: mode}
+             opts
            ) do
         {:ok, _pid} ->
-          Logger.notice("Proxy started #{mode} on port #{port}")
+          Logger.notice("Proxy started #{opts.mode}(local=#{opts.local}) on port #{port}")
```
abc3: Here we need to get a value from `ranch` by id, because the `port` is `0`. Although, it might be better to start the range from a defined value to make this behavior more deterministic
```suggestion
          port = :ranch.get_port(key)
          Logger.notice("Proxy started #{opts.mode}(local=#{opts.local}) on port #{port}")
```

v0idpwn: Added the port!

I'd argue that starting on a range may be "less deterministic". Here's the behaviour is more predictable: we listen on `0`, and the OS gives us a port that will always be free. If we try to determine it ourselves, we might get conflicts, etc.

abc3: we can choose something that we are sure will be free, for example, starting at 45_000

abc3: and we will always know that these ports are used by local proxies

v0idpwn: There's no such thing as surely free ports, specially in the ephemeral range :P

abc3: well, there are a bunch of strict "safe" examples that supavisor uses: 4000, session's, transaction's, etc 😜

v0idpwn: But these aren't in the ephemeral range...

v0idpwn: We could maybe pick something outside it though (like 12_000 or smth) 🤔

PR #709, opened 2025-07-22
feat: configure internal proxy ports via explicit lists

abc3:


config/test.exs
```diff
@@ -18,7 +18,8 @@ config :supavisor,
   metrics_blocklist: [],
   node_host: System.get_env("NODE_IP", "127.0.0.1"),
   availability_zone: System.get_env("AVAILABILITY_ZONE"),
-  local_proxy_shards: System.get_env("LOCAL_PROXY_SHARDS", "4") |> String.to_integer(),
+  session_proxy_ports: System.get_env("SESSION_PROXY_PORTS", "12100,12101,12102,12103") |> String.split(",") |> Enum.map(&String.to_integer/1),
+  transaction_proxy_ports: System.get_env("TRANSACTION_PROXY_PORTS", "12104,12105,12106,12107") |> String.split(",") |> Enum.map(&String.to_integer/1),
```
v0idpwn: Probably need to update the makefile so that node2 doesn't try using the same ports, etc

abc3: ah, right 🤦

v0idpwn (approved):

v0idpwn:
Thank you! 🎉

PR #711, opened 2025-07-23
feat: more consistency with postgres errors on auth errors

### LINKED ISSUE #710: SCRAM authentication correctness/improvements
# Bug report

## Describe the bug

Currently, Supavisor returns SCRAM errors in the `scram-server-final` message error field. The message is an **arbitrary string**. `psql`, `Postgrex`, and some other clients handle this gracefully and show this error to the user. 

Some other clients have different behaviour:

- node-postgres ignores this error completely (which is probably also wrong) and errors due to the `v` field not being present.
- pgjdbc (apparently) only accepts the server errors defined in the RFC: https://github.com/ongres/scram/blob/dde95f22d0876d9d6478a75ffc57353c1c31d7a1/scram-common/src/main/java/com/ongres/scram/common/exception/ServerErrorValue.java#L2

Postgres leverages the fact that SCRAM allows you to return non-scram errors instead of returning `scram-server-final`, to ensure consistency across different auth methods and return "postgres errors" instead of "scram errors", see the comment: https://github.com/postgres/postgres/blob/2047ad068139f0b8c6da73d0b845ca9ba30fb33d/src/backend/libpq/auth-scram.c#L423.

## Expected behavior

Align our behaviour with Postgres' to improve the auth errors saw by clients.

v0idpwn:
Fixes #710 (see the issue for a more thorough description).


lib/supavisor/client_handler.ex
```diff
@@ -351,14 +351,11 @@ defmodule Supavisor.ClientHandler do
           "ClientHandler: Exchange error: #{inspect(reason)} when method #{inspect(method)}"
         )
 
-        msg =
-          if method == :auth_query_md5,
-            do: Server.error_message("XX000", reason),
-            else: Server.exchange_message(:final, "e=#{reason}")
+        msg = postgresql_auth_error(reason, data.user)
```
abc3: maybe move this line closer to `HandlerHelpers.sock_send(sock, msg)`

chasers (approved):

PR #718, opened 2025-08-01
chore: fix nix development environment missing libclang

nelsonmestevao:

Explicitly list `libclang` dependency for the flake setup and configure the `LIBCLANG_PATH` variable to use that dependency for development using nix flake environment. I originally fixed it using `llvmPackages_16.libclang`, but it also works with `pkgs.libclang` so I am changing to this one in the last pushed commit. 

image

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -130,7 +130,7 @@
                   pkgs.protobuf
                   pkgs.cargo-outdated
                   pkgs.prom2json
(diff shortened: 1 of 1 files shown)


v0idpwn (approved):

v0idpwn:
Thank you!

v0idpwn:
Thank you!

PR #720, opened 2025-08-04
[WIP] test: add tests for different auth methods (scram-sha-256, password, md5)

laraujo7:

Introduces tests for tenants using different authentication methods (scram-sha-256, password, md5). 
:warning:  MD5 still needs fixing

## Additional context

The new docker compose creates three instances, one for each of the authentication methods. Then existing tests regarding tenants are run against each of them.


config/test.exs
```diff
@@ -41,6 +41,29 @@ config :supavisor, Supavisor.Repo,
   pool_size: 10,
   port: 6432
 
+config :supavisor, :tenants,
```
v0idpwn: I think we shouldn't use`:tenants` as the name here as we already have a different "tenant" concept in the application. Let's avoid loading the term too much.

priv/repo/seeds_before_migration.exs
```diff
@@ -1,6 +1,22 @@
...
+defmodule PasswordRepo do
+  use Ecto.Repo,
+    otp_app: :supavisor,
+    adapter: Ecto.Adapters.Postgres
+end
+
+PasswordRepo.start_link(tenants_conf[:password])
+
+[MD5Repo, PasswordRepo, Supavisor.Repo]
+|> Enum.each(fn repo -> query(repo, "create schema if not exists _supavisor", []) end)
```
v0idpwn: I feel like we shouldn't need these extra repos (?), since they are our clients/tenants, we should only need a repo for our own supavisor database?

test/supavisor/monitoring/tenant_test.exs
```diff
@@ -35,37 +35,40 @@ defmodule Supavisor.PromEx.Plugins.TenantTest do
     end
   end
 
-  describe "execute_client_connections_lifetime" do
-    setup ctx do
-      create_instance([__MODULE__, ctx.line])
-    end
+  # for {key, port} <- [{"scram-sha-256", 6432}, {"md5", 6433},{"password", 6434}] do
+  for {key, port} <- [{"scram-sha-256", 6432}, {"password", 6434}] do
```
v0idpwn: Should probably have a function "wrapping" that list, and potentially reutrn all the connection options (hostname, port, database, etc) then we can merge the whole thing into the `ctx`, will probably make it smoother to add it in more places/will need less ad-hoc changes.

priv/repo/seeds_before_migration.exs
laraujo7: Yeah, you are right we dont :muscle:

v0idpwn (approved):

PR #725, opened 2025-08-12
Etienne/sec 488 support jit access in supavisor

staaldraad:

Feature

Introduces support for Just-in-time (JIT) access via a Personal Access Token (PAT) or JWT, which is validated against a remote API. This is implemented as a separate tenant feature, since it is not a standard postgres feature. The upstream tenant must also be configured for JIT, which requires a PAM to be installed and configured in the pg_hba. 

With JIT in use, a disconnect happens between what the server expects for auth (pg_hba.conf) and what `pg_authid` stores for user auth. In most cases pg_authid will have the credentials saved as scram-sha-256, but with the use of PAM for authentication, the database uses `AuthenticationCleartext_password`. For this reason, the Supavisor must be made aware of the tenants changed auth, and we use a new `use_jit` configuration value for the tenant. When this is active, the pooler will use AuthenticationCleartext_password, and support either logging in with the valid user password (which will be checked against the scram-sha-256 retrieved from pg_authid) or a PAT/JWT that is validated against the upstream API server configured via `jit_api_url`. 

Cached credentials keep working as before, except in the case of the PAT/JWT, the check is always performed against the upstream API server. This ensures that any revocation of JIT access is respected, alongside the expiration/revocation of the auth tokens, something that happens outside of the database.

Doesn't support JIT.

## What is the new behavior?

Supports JIT, allowing login with:

JWT:
```
psql 'postgresql://postgres.dev_tenant2:eyJhbGciOiJSUzI1NiIsImtpZCI6IjcyYjY2NjA1IiwidHlwIjoiSldUIn0.eyJhYWwiOiJhYWwyIiwiYW1yIjpbeyJtZXRob2QiOiJ0b3RwIiwidGltZXN0YW1wIjOnt9fQ.XbOq_XWg@localhost:6543/postgres'
```

PAT:
```
psql 'postgresql://postgres.dev_tenant2:sbp_39wBdIEXAMPLESdMIg@localhost:6543/postgres'
```

Password:

```
psql 'postgresql://postgres.dev_tenant2:ExAmPlePA$$@localhost:6543/postgres'
```

## Additional context

In the case of JWT/PAT, there is double lookup to the API that happens on first connection (new pool). First by pooler, and then by the upstream tenant. After that, the API lookup only happens in pooler for every new attempt to join the pool with a PAT/JWT.

The API server is expected to respond with 200 OK and a list of roles the token is allowed to assume, or return permission denied if the user is not authorized for JIT or there are any upstream errors.

https://github.com/user-attachments/assets/7e9b81f9-c3ec-4bfc-b24b-b634fcd64a4d


lib/supavisor/helpers.ex
```diff
@@ -399,4 +399,79 @@ defmodule Supavisor.Helpers do
...
+
+  ## Parameters
+
+  - `token`: The token string
+  """
+  def token_matches?(<<"sbp_", _rest::binary>>), do: true
+
+  def token_matches?(token) do
+    case String.split(token, ".") do
+      [h, p, _sig] when binary_part(h, 0, 3) == "eyJ" and binary_part(p, 0, 3) == "eyJ" ->
```
v0idpwn:
```suggestion
      ["eyJ" <> _, "eyJ" <> _, _sig] ->
```

staaldraad:
@v0idpwn 👋🏼 I've come back to this project. The PR has been rebased and updated to work with the latest changes, what do you need from me to move this forward?

lib/supavisor/client_handler.ex
```diff
@@ -613,6 +620,33 @@ defmodule Supavisor.ClientHandler do
     end
   end
 
+  def handle_event(:info, {proto, socket, bin}, :auth_password_wait, data)
+      when proto in @proto do
+    auth_context = data.auth_context
+    {:ok, {ip, _port}} = :inet.peername(socket)
+
```
v0idpwn: I think this won't work with ssl sockets, but we have a function in the codebase that handles both

lib/supavisor/helpers.ex
```diff
@@ -435,4 +438,74 @@ defmodule Supavisor.Helpers do
...
+
+  def token_matches?(token) do
+    case String.split(token, ".") do
+      ["eyJ" <> _, "eyJ" <> _, _sig] ->
+        true
+
+      _ ->
+        false
+    end
+  end
```
v0idpwn, line 476: This won't be needed anymore, right?

staaldraad: no, it is still needed for https://github.com/supabase/supavisor/pull/725/changes#diff-564f23a9c1bf49d69e012f651416b125725ee3938d92c9e855bbc40ada3c2063R91

Otherwise we'd send user passwords to the API. We only want to send JWT or PATs

v0idpwn: >Otherwise we'd send user passwords to the API. We only want to send JWT or PATs

I thought we would make it based in the user, i.e.: if the user is `foo.jit` then we assume its JWT/PAT, otherwise assume its a password.

staaldraad: unfortunately it doesn't work end-to-end. 
the problem lies with authentication on the postgres side, it can either be 
```
host all all 0.0.0.0/0 pam
```
or 
```
host all all 0.0.0.0/0 scram-sha256
```
But it can't be both and it also doesn't allow fall-through. It also doesn't support the option of 
```
host user.jit all 0.0.0.0/0 pam
host user all 0.0.0.0/0 scram-sha256
```
because it will expect `user.jit` to be an actual user/role in the database.

Using `user.jit.tenant` works for the case of jit only, 

Image

But in the normal flow it would only work for the first part, authenticating with supavisor. The postgres database and supavisor would be out of sync after that, with the database requesting a password, while supavisor can only supply the scram-sha256.

Image

We would need to modify postgres to have the same logic to request password for `user.jit` and expect scram-sha256 for `user`

staaldraad: FWIW, I discovered during this that for a similar reason md5 auth never actually worked and can't work. The protocol is correct in supavisor, but because of `user.tenant` the md5 hash will never be correct. the client does 
```
hash1 = crypto.md5(password, user)
hash2 = "md5" <> crypto.md5(hash1, salt)
```
the hash that supavisor uses is the one from the database, 
```
"md5" <> crypto.md5(password, user)
```
which is fine and we could combine with the salt. The problem is the client will always use a different value for `user` from the one that is in the database hash... the client will use `user.tenant` while the database will always just have `user` :(

PR #728, opened 2025-08-19
fix(authn): when credential changes are detected, retry the auth request 

v0idpwn:
- Fixes an issue where the first authn request after a credential change always would fail.
- Adds a cache refresh limit of 3 per minute per pool (triggered every time the password is wrong). Previosuly, limited to once every 15s. This is mostly due to tests. In tests, if we don't have this, the hard limit of once per 15s causes the test to flak.
- Adds an optimization where we use the existing SecretsChecker connection instead of opening a new one to refresh the cache.

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -302,7 +302,7 @@ defmodule Supavisor.ClientHandler do
 
             key = {:secrets, tenant_or_alias, user}
 
-            case auth_secrets(info, user, key, :timer.hours(24)) do
+            case cached_get_secrets(data.id, info, user, key, :timer.hours(24)) do
               {:ok, auth_secrets} ->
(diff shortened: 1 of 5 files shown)


lib/supavisor/client_handler.ex
```diff
@@ -712,6 +669,112 @@ defmodule Supavisor.ClientHandler do
...
+        retry_with_get_secrets(sock, {method, secrets}, data, info, key)
+      end
+    else
+      Logger.debug("ClientHandler: No retry condition met")
+      :no_retry
+    end
+  end
+
+  defp retry_with_cached_secrets(sock, {method, secrets}, data, info, key) do
+    case cached_get_secrets(data.id, info, data.user, key, 15000) do
```
v0idpwn:
```suggestion
    case cached_get_secrets(data.id, info, data.user, key, 15_000) do
```

PR #730, opened 2025-08-25
fix: always terminate db_handler when owner client_handler terminates

v0idpwn:
Among other things, this patch addresses a race condition when the socket was closed while DbHandler had `active: false`. This led to the DbHandler not receiving the TCP closed message, and the client handler crashing due to it. The DbHandler would stay on the pool with the dead socket, and and successive checkouts would get an unusable connection.

This patch ensures that the DbHandler is closed whenever the owner terminates, regardless of how it terminates or pool mode. This may cause the transaction pooler to restart connections more often, but should give us more correct behaviour.

### DIFF
--- a/lib/supavisor/db_handler.ex
+++ b/lib/supavisor/db_handler.ex
@@ -360,20 +360,16 @@ defmodule Supavisor.DbHandler do
   end
 
   # linked client_handler went down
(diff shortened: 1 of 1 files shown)


chasers (approved):

v0idpwn:
Closing this as its part of #694, but it was backported to the v2.1.x branch.

PR #731, opened 2025-08-26
fix: correct tenant cache retrieval and add tests

laraujo7:

Bug fix and tests

- `get_tenant_cache/2` was not working correctly in some scenarios  
- Missing tests for tenant/cluster lookups and cache behavior  

## What is the new behavior?

- Fixed `get_tenant_cache/2` to correctly return tenants and handle cache expiration  
- Added tests for:  
  - `list_tenants/0` returns all tenants  
  - `get_tenant_by_external_id/1`  
  - `get_cluster_by_alias/1`  
  - `get_tenant_cache/2` (fetching from cache and after expiry)


lib/supavisor/tenants.ex
```diff
@@ -57,14 +68,14 @@ defmodule Supavisor.Tenants do
...
+      {:ok, {:cached, value}} -> value
+      {:commit, {:cached, value}, _} -> value
     end
   end
 
   @spec get_tenant(String.t() | nil, String.t() | nil) :: Tenant.t() | nil
-  def get_tenant(nil, sni) when sni != nil do
-    Tenant |> Repo.get_by(sni: sni)
+  def get_tenant(nil, sni_hostname) when sni_hostname != nil do
+    Tenant |> Repo.get_by(sni_hostname: sni_hostname)
```
v0idpwn, line 78: Looks like this was broken (?)
Is it even used? Should it be removed?

PR #734, opened 2025-08-28
fix: correct get_user_cache fuction and add tests

laraujo7:

- Bug fix  
- Test improvements  

## What is the new behavior?

- Renamed `get_user_cache/4` to `fetch_user_cache/4` to better reflect its return value (`{:ok, ...} | {:error, ...}`).  
- Fixed behavior where errors were being cached — these are now ignored instead of stored.  
- Added missing test coverage for both `:single` and `:cluster` modes of `fetch_user/4`.  

## Additional context

- Improves clarity of function semantics.  
- Ensures the cache only stores valid user lookups.  
- Strengthens confidence in tenant and user resolution logic.


lib/supavisor/tenants.ex
```diff
@@ -73,46 +73,43 @@ defmodule Supavisor.Tenants do
...
-      [%ClusterTenants{} = ct] ->
-        get_user(:single, user, ct.tenant_external_id, sni_hostname)
-
-      [_ | _] ->
-        {:error, :multiple_results}
+    case Repo.one(query) do
+      %ClusterTenants{} = ct ->
+        fetch_user(:single, user, ct.tenant_external_id, sni_hostname)
 
       _ ->
```
nelsonmestevao:
```suggestion
      nil ->
```

PR #737, opened 2025-09-04
chore: add handle db authentication_sasl_password and authentication_server_first_message tests

laraujo7:

- Adds tests for handling `:authentication_sasl_password` messages
- Adds tests for handling `:authentication_server_first_message`


test/supavisor/db_handler_test.exs
```diff
@@ -166,6 +166,84 @@ defmodule Supavisor.DbHandlerTest do
...
+
+      %{data: data, b: b}
+    end
+
+    test "handles SASL authentication and sets nonce", %{data: data, b: b} do
+      # `82` is `?R`, which identifies the payload tag as `:authentication`
+      # `0, 0, 0, 22` is the packet length
+      # `0, 0, 0, 10` is the authentication type, identified as `:authentication_sasl_password`
+      payload = "SCRAM-SHA-256" <> <<0>>
+      bin = <<82, 0, 0, 0, 22, 0, 0, 0, 10, payload::binary>>
```
v0idpwn:
```suggestion
      bin = <<?R, 22::32,  10::32, "SCRAM-SHA-256", 0>>
```
More readable, same for other instances (feel free to change other tests that have this pattern too in the file)

test/supavisor/db_handler_test.exs
```diff
@@ -166,6 +166,84 @@ defmodule Supavisor.DbHandlerTest do
...
+          secrets: fn -> %{user: "user", password: "pass"} end
+        },
+        sock: {:gen_tcp, a},
+        nonce: "some nonce"
+      }
+
+      %{data: data, b: b}
+    end
+
+    test "handles SASL authentication and sets nonce", %{data: data, b: b} do
```
v0idpwn: not a fan of having something named `b`. It's fine when its used in a single scope but not something you want to pass in setup to many tests

test/supavisor/db_handler_test.exs
```diff
@@ -166,6 +166,84 @@ defmodule Supavisor.DbHandlerTest do
...
+      content = {:tcp, b, bin}
+
+      assert {:keep_state, %{nonce: nil}} = Db.handle_event(:info, content, :authentication, data)
+    end
+  end
+
+  describe "handle_event/4 info tcp authentication authentication_server_first_message payload events" do
+    test "handles server first message" do
+      server_first = "r=nonce12345nonce67890,s=c2FsdA==,i=4096" <> <<0>>
+      pkt_len = 4 + byte_size(server_first)
```
v0idpwn: Shouldn't it be 8 + byte_size(server_first)?

v0idpwn:
>Does validating that nonce and server_proof is different enough for the test?
I'd prefer if we could actually test that we get the nonce and server_proof from the database, but if it's too hard its fine like this

test/supavisor/db_handler_test.exs
laraujo7: Yeap. I was following the previous tests but I really prefer like that.

laraujo7:
> > Does validating that nonce and server_proof is different enough for the test?
> 
> I'd prefer if we could actually test that we get the nonce and server_proof from the upstream db, but if it's too hard its fine like this

That would be perfect, but I think that nonce is a random value.
`nonce = :pgo_scram.get_nonce(16)`
"Resulting string length isn't guaranteed, but it's guaranteed to be unique and  contain `NumRandomBytes' of random data."

test/supavisor/db_handler_test.exs
laraujo7: Yes, that was totally my bad. Fixed :+1:

v0idpwn (approved):

PR #744, opened 2025-09-17
fix(defrag): improve statems, caching, logs, circuit breaking

v0idpwn:
This is a big "defrag" PR solving many of the paper cuts from the current architecture:

### Secret caching
Ensure we don't store secrets inside start options, which caused unintentional caching on the pool, and made it required to restart the pool if a password changed.

Introduces a central secret caching module instead of relying on arbitrary cachex calls.

Breaks into two types of secrets:
- validation secrets: used to validate credentials on connection attempts, **can be skipped for users where we don't want caching, e.g.: users with single use passwords**
- upstream secrets: used for connection. 

### Logs
- Reduces log duplication: terminate processes normally on expected situations
- Better messages

### Pool termination
- Terminate pool early if Db doesn't exist anymore: this resolves an issue where we would keep getting spammed by prisma shadow dbs that get deleted but we would keep trying to connect to them.

### Circuit breaker
- Under situations we can't proceed (can't connect to the database, can't get secrets) don't allow pool initialization. Give a user friendly message.

### ClientHandler changes
- Break ClientHandler into more modules
- Break ClientHandler into more states, with no arbitrary socket receives
- Try to isolate some parts of ClientHandler: error handling, auth, etc
- Remove dead code (read replicas)
- Some refactoring/renaming, e.g.: db_pid -> db_connection (it was a tuple)
- Use state enter for state tracking in ClientHandler (removes manual logger.metadata calls, adds telemetry)
- Improve active counting

### Breaking changes
Support for aliases removed

### Misc
- Better error decoding and formatting


chasers (approved):

lib/supavisor/db_handler.ex
```diff
@@ -188,14 +233,36 @@ defmodule Supavisor.DbHandler do
...
         handle_authentication_error(data, reason)
-        Logger.error("DbHandler: Auth error #{inspect(reason)}")
+        Logger.error("DbHandler: Auth error #{inspect(error)}")
         {:stop, :invalid_password, data}
 
+      {:error_response, %{"S" => "FATAL", "C" => "3D000"} = error} ->
+        Logger.error("DbHandler: Database does not exist: #{inspect(error)}")
+        encode_and_forward_error(error, data)
+
+        unless data.proxy do
```
abc3: unless is deprecated

lib/supavisor/secret_cache.ex
```diff
@@ -0,0 +1,205 @@
...
+
+  @default_secrets_ttl :timer.hours(24)
+
+  @doc """
+  Gets validation secrets for validating incoming client authentication requests.
+  """
+  def get_validation_secrets(tenant, user) do
+    Logger.info("get_validation_secrets(#{tenant}, #{user})")
+
+    case Cachex.get(Supavisor.Cache, {:secrets_for_validation, tenant, user}) do
```
abc3, line 29: these new keys probably need to be added to Supavisor.del_all_cache

v0idpwn: Good call, will push

PR #751, opened 2025-10-06
feat: add endpoint to change user credentials

v0idpwn:
When using `auth_query` for authn, we should be able to change the credentials of the managing user without a pool shutdown/downtime. This adds a new endpoint with this functionality.


lib/supavisor.ex
```diff
@@ -136,6 +136,47 @@ defmodule Supavisor do
...
+    |> Enum.reduce(%{}, fn {_pid,
+                            %{
+                              user: user,
+                              mode: mode,
+                              type: type,
+                              db_name: db_name,
+                              search_path: search_path
+                            }},
+                           acc ->
+      id = {{type, tenant}, user, mode, db_name, search_path}
```
chasers, line 168: Feel like we should have a fn for this somewhere else?

chasers (approved):
But, otherwise looks good

lib/supavisor.ex
v0idpwn: Also thought but didn't find any other occurrence of the same pattern 🤔

lib/supavisor/secret_checker.ex
```diff
@@ -39,27 +39,49 @@ defmodule Supavisor.SecretChecker do
...
+  def handle_continue(:init_conn, state) do
+    # Secret Checker always uses the manager user
+    tenant =
+      state.tenant_external_id
+      |> Tenants.get_tenant_by_external_id()
+
+    auth_query_user =
+      case tenant.users do
+        [u] -> u
+        users -> Enum.find(users, fn u -> u.is_manager end)
```
abc3: might be cleaner with the capture operator

```suggestion
        users -> Enum.find(users, & &1.is_manager)
```

lib/supavisor/secret_checker.ex
```diff
@@ -39,27 +39,49 @@ defmodule Supavisor.SecretChecker do
...
   end
 
-  def handle_continue(:init_conn, %{auth: auth} = state) do
-    {:ok, conn} = start_postgrex_connection(auth)
-    {:noreply, %{state | conn: conn}}
+  def handle_continue(:init_conn, state) do
+    # Secret Checker always uses the manager user
+    tenant =
+      state.tenant_external_id
+      |> Tenants.get_tenant_by_external_id()
```
abc3:
```suggestion
    tenant = Tenants.get_tenant_by_external_id(state.tenant_external_id)
```

lib/supavisor/secret_checker.ex
```diff
@@ -73,7 +95,7 @@ defmodule Supavisor.SecretChecker do
   end
 
   def check(interval \\ @interval),
-    do: Process.send_after(self(), :check, interval)
+    do: Process.send_after(self(), :check, interval + jitter())
```
abc3, line 103: does it really need it? if i'm not mistaken, it adds 1–1_500 ms of random jitter, which is negligible compared to a 15_000 ms interval

lib/supavisor.ex
```diff
@@ -470,4 +443,45 @@ defmodule Supavisor do
...
+        %{
+          db_user: db_user,
+          db_password: db_pass,
+          db_user_alias: alias
+        }
+      ]
+    } = tenant_record
+
+    %{
+      host: String.to_charlist(db_host),
```
abc3:
```suggestion
      host: to_charlist(db_host),
```

lib/supavisor.ex
```diff
@@ -470,4 +443,45 @@ defmodule Supavisor do
...
+      sni_hostname: if(sni_hostname != nil, do: to_charlist(sni_hostname)),
+      port: db_port,
+      user: db_user,
+      alias: alias,
+      auth_query: auth_query,
+      database: if(db_name != nil, do: db_name, else: db_database),
+      password: fn -> db_pass end,
+      application_name: "Supavisor",
```
abc3: it would be nice to keep the possibility to identify the connection from the pg side
```suggestion
      application_name: "Supavisor (auth query)",
```

lib/supavisor/secret_checker.ex
v0idpwn: It's just that when we start a node, we tend to get a lot of pools initialized around the same second. So if we have a fixed interval, it will be "spiky" every 15s, and if we add a jitter it mitigates this issue. Not that it is a big problem, just something I usually like to do.

lib/supavisor.ex
v0idpwn: Added an extra parameter there

lib/supavisor/secret_checker.ex
v0idpwn: I'm a `fn` enjoyer, but thats fair

lib/supavisor/secret_checker.ex
abc3: it makes sense!

PR #760, opened 2025-10-24
fix: don't stop the pool if config change request didn't change anything

v0idpwn:


lib/supavisor/tenants.ex
```diff
@@ -209,6 +209,16 @@ defmodule Supavisor.Tenants do
     %Tenant{}
     |> Tenant.changeset(attrs)
     |> Repo.insert()
+    |> case do
+      {:ok, tenant} ->
+        # Clear cache after creating tenant to ensure cached "not found" results are invalidated
+        cleanup_result = Supavisor.del_all_cache_dist(tenant.external_id)
```
abc3, line 215: maybe this is a good time to make `del_all_cache_dist` non gradual

lib/supavisor/tenants.ex
```diff
@@ -229,6 +239,48 @@ defmodule Supavisor.Tenants do
...
+    # Some fields may not require pool termination.
+    if Enum.empty?(changeset.changes) do
+      Logger.info(
+        "No changes detected for tenant #{tenant.external_id}, skipping update and restart"
+      )
+
+      {:ok, tenant}
+    else
+      # Clear cache before update
+      cleanup_result = Supavisor.del_all_cache_dist(tenant.external_id)
```
abc3, line 269: why clear cache before update?

PR #766, opened 2025-10-29
chore: deploy soft use11

v0idpwn:


deploy/upgrade/prod.config
```diff
@@ -1,18 +1,10 @@
 [
-    {upgrade_from, "2.7.1"},
+    {upgrade_from, "2.7.3"},
```
v0idpwn, line 2: Notice this one has 2.7.3 as base, as most nodes were started at 2.7.3

PR #780, opened 2025-11-27
fix: bump Peep for dialyzer fix, remove 404 for no metrics tenants

v0idpwn:
Bumps Peep for a typespec fix.

Also removes 404 status code for empty tenant metrics. It didn't work with Peep 4.2 (due to Peep 4.2 always returning at least an "EOF" comment for OpenMetrics compatibility). It also __probably__ makes more sense to return an empty 200 in this situation: 404 would imply the endpoint is invalid, but the tenant might just not have any recent metrics.


mix.exs
```diff
@@ -53,7 +53,7 @@ defmodule Supavisor.MixProject do
       {:phoenix_live_view, "~> 1.0"},
       {:phoenix_live_dashboard, "~> 0.7"},
       {:telemetry_poller, "~> 1.0"},
-      {:peep, "~> 4.2", override: true},
+      {:peep, "~> 4.2 and >= 4.2.1", override: true},
```
abc3, line 56: maybe just specify 4.2.1?
```suggestion
      {:peep, "4.2.1", override: true},
```

v0idpwn: 4.2.1 won't accept 4.3, this one does

abc3: ahh, got it

PR #783, opened 2025-12-03
fix: potential zombie client handlers on fatal tls alerts

v0idpwn:
When we get fatal alerts, the TLS connection is closed, but we don't get a `ssl_closed` messsage. This handles the alert, checks if it's fatal and if yes, we terminate. Keeps the previous behaviour for warning alerts or other TLS errors.

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -482,11 +482,19 @@ defmodule Supavisor.ClientHandler do
   # For FATAL alerts, Erlang doesn't send ssl_closed, so we must terminate here.
   # The alert level is only available by parsing the message string ("Warning - " or "Fatal - ").
   def handle_event(:info, {:ssl_error, sock, {:tls_alert, {_reason, msg}}}, _, %{sock: {_, sock}}) do
-    if String.contains?(msg, "Fatal -") do
-      Logger.warning("ClientHandler: Received fatal TLS alert: #{msg}, terminating connection")
+    msg_string = to_string(msg)
+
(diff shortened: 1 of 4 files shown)


lib/supavisor/client_handler.ex
```diff
@@ -478,6 +478,19 @@ defmodule Supavisor.ClientHandler do
...
+  # For FATAL alerts, Erlang doesn't send ssl_closed, so we must terminate here.
+  # The alert level is only available by parsing the message string ("Warning - " or "Fatal - ").
+  def handle_event(:info, {:ssl_error, sock, {:tls_alert, {_reason, msg}}}, _, %{sock: {_, sock}}) do
+    if String.contains?(msg, "Fatal -") do
+      Logger.warning("ClientHandler: Received fatal TLS alert: #{msg}, terminating connection")
+      {:stop, :normal}
+    else
+      Logger.warning("ClientHandler: Received TLS warning alert: #{msg}, keeping connection alive")
+      :keep_state_and_data
+    end
```
v0idpwn: I don't like the `String.contains?/2` here, but Erlang doesn't send us the level outside of the message.

The same reason may have different levels, so we can't use the `reason`. For example, we see `user_canceled` with `WARNING` and `FATAL` in production.

v0idpwn: cc @josevalim the atoms are there but the level is, in some cases, arbitrary. The same one can be warning or fatal depending on whether the implementation decides whether its recoverable or not (?). Unsure why it is like that, but it is.

PR #804, opened 2025-12-08
chore: deploy v2.8.6 in remaining cluster idx1

v0idpwn:


deploy/upgrade/prod.config
```diff
@@ -4,10 +4,12 @@
     {upgrade_to, "2.8.5"},
```
v0idpwn:
```suggestion
    {upgrade_to, "2.8.6"},
```

chasers (approved):

PR #811, opened 2025-12-15
fix: only start ranch listeners after infra is set up

v0idpwn:
Currently, the ranch listeners are started very early, which means we accept connections before being healthy. This fixes the issue by starting the ranch listeners later.

This makes Supavisor fail at startup if these ports are busy (instead of just logging), which caused some tests to need fixes.

### DIFF
--- a/test/supavisor/logger/logflare_formatter_test.exs
+++ b/test/supavisor/logger/logflare_formatter_test.exs
@@ -11,20 +11,20 @@ defmodule Supavisor.Logger.LogflareFormatterTest do
   defmodule FakeLogger do
     @behaviour :logger_handler
 
-    def install(id, opts) do
-      pid = start_supervised!({Agent, fn -> [] end})
+    def install(id, opts, filter_pid) do
+      agent_pid = start_supervised!({Agent, fn -> [] end})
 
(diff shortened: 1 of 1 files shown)


chasers (approved):

test/integration/cluster_pooling_test.exs
```diff
@@ -5,13 +5,35 @@ defmodule Supavisor.Integration.ClusterPoolingTest do
...
+      Cluster.start_node_unclustered(
+        :test_node_2,
+        %PortConfig{
+          proxy_port_transaction: 7661,
+          proxy_port_session: 7662,
+          proxy_port: 7663,
+          session_proxy_ports: [14_100, 14_101, 14_102, 14_103],
+          transaction_proxy_ports: [14_104, 14_105, 14_106, 14_107]
+        }
+      )
```
mentels, line 36: @v0idpwn was this change required or you just wanted make different nodes make use of different port?

v0idpwn: It was required, because now we actually crash if the ports are busy, so the nodes would use the same port by default and crash.

PR #824, opened 2025-12-23
fix: monitor TCP socket and terminate ClientHandler if it closed

v0idpwn:
In at least one case, our current handling didn't catch the closed socket, leading to potential zombie connections. This changes the approach according to the discussion on #783, instead relying on monitoring. 

The most complex part is the upgrade process: we need to ensure the existing ClientHandlers are updated to monitor their sockets, but since some of the sockets were already upgraded to TLS, we only have the `sslsocket` record. In these cases, we monitor the `connection_handler` process. For this reason, we must be able to handle both ports and processes.

In case the change is reverted, no `code_change` is necessary, since the DOWN message will be caught by the catch all clause.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -8,6 +8,11 @@ parse_integer_list = fn numbers when is_binary(numbers) ->
   |> Enum.map(&String.to_integer/1)
 end
 
+db_socket_options =
+  if System.get_env("SUPAVISOR_DB_IP_VERSION") == "ipv6",
+    do: [:inet6],
+    else: [:inet]
+
 secret_key_base =
   if config_env() in [:dev, :test] do
     "3S1V5RyqQcuPrMVuR4BjH9XBayridj56JA0EE6wYidTEc6H84KSFY6urVX7GfOhK"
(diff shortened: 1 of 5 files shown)


lib/supavisor/client_handler.ex
```diff
@@ -9,6 +9,14 @@ defmodule Supavisor.ClientHandler do
 
   require Logger
 
+  require Record
+
+  # Import the sslsocket record definition
```
v0idpwn, line 14:
```suggestion
```

relups/2.8.8-2.8.9/appups/supavisor-2.8.9.appup
```diff
@@ -0,0 +1,17 @@
+{"2.8.9",
+ [{"2.8.8",
+   [{apply,{'Elixir.Supavisor.HotUpgrade',apply_runtime_config,["2.8.9"]}},
+    {apply,{'Elixir.Supavisor.HotUpgrade',reint_funs,[]}},
+    {update,'Elixir.Cluster.Strategy.Postgres',{advanced,[]},[]},
```
mentels, line 5: Is it a common pattern to always execute these 3 instructions on hot code reload?

v0idpwn: These are generated by `Supavisor.HotUpgrade` module, and __generally yes__.

`apply_runtime_config` reloads the config in case we touched `runtime.exs`, does nothing otherwise
`reint_funs` reinitializes secret anonymous functions that are in ETS (meaning they start referencing the HotUpgrade` module not their original module), so they don't become invalid when we reload other modules.

PR #831, opened 2026-01-12
fix: support IPv6 in Postgres topology as well

tvogel:

Bug fix: Supavisor failed to start up when having an IPv6-only Postgres host, e.g. on fly.io's internal network.

After startup, supavisor will not be functional and loop error output like this:
```
2026-01-12T16:38:56Z app[1853305f329658] fra [info]16:38:56.216 region=local [error] #PID<0.4921.0> (Postgrex.Notifications) failed to connect to Postgres: ** (DBConnection.ConnectionError) tcp connect (example-db.internal:5432): non-existing domain - :nxdomain
```

## What is the new behavior?

Supavisor starts up successfully and provides connectivity to the postgres database.

## Additional context

(This was fixed as part of working on https://github.com/SupaAutoFly/SupaAutoFly to support the latest supabase self-hosted release.)


v0idpwn:
Thanks for the PR, looks great!

tvogel:
Awesome, thanks for the merge!

PR #837, opened 2026-01-19
perf(session): reuse backend connections when possible

v0idpwn:
When terminating gracefully and not in transaction, the ClientHandler can run a cleanup in the DbHandler and put it back to the pool. This reduces the overhead of opening new connections in the session mode, making supavisor a more performant session pooler, akin to PgBouncer.


lib/supavisor/client_handler.ex
```diff
@@ -795,6 +797,23 @@ defmodule Supavisor.ClientHandler do
     Logger.log(level, "ClientHandler: terminating with reason #{inspect(reason)}")
   end
 
+  defp maybe_cleanup_db_handler(state, data) do
+    if state == :idle and data.mode == :session and data.db_connection != nil and
+         !Supavisor.Helpers.no_warm_pool_user?(data.user) do
+      Logger.debug("ClientHandler: Performing session cleanup before termination")
+      {pool, db_pid, _} = data.db_connection
+
+      case DbHandler.attempt_cleanup(db_pid) do
```
mentels, line 811: I wonder if it wasn't be safer to have a timeout here as well.

Can we rule out a scenario in wich an "unrelated" event arrives to the `DbHandler` while in `:waiting_cleanup` and cancels the `DbHandler` `:cleanup_timeout`?

lib/supavisor/db_handler.ex
```diff
@@ -244,6 +252,11 @@ defmodule Supavisor.DbHandler do
     {:keep_state, %{data | reconnect_retries: retry + 1}, {:next_event, :internal, :connect}}
   end
 
+  def handle_event(:timeout, :cleanup_timeout, :waiting_cleanup, _data) do
+    Logger.error("DbHandler: Cleanup timeout, shutting down")
+    {:stop, :normal}
+  end
+
```
mentels: As stated in the comment above,  I'm thinking that we might need a catch-all clause for the `:waiting_cleanup` so that we eventually stop the DbHandler if were're not done with the clean-up.

Otherwise, unless I misread thing, any event that will be handled by the catch-all handle event will cancel the timeout.

lib/supavisor/db_handler.ex
```diff
@@ -347,6 +360,24 @@ defmodule Supavisor.DbHandler do
...
+  def handle_event(:info, {proto, _, bin}, :waiting_cleanup, %{caller: caller} = data)
+      when is_pid(caller) and proto in @proto do
+    buffered_bin = data.pending_bin <> bin
+
+    cond do
+      String.ends_with?(buffered_bin, Server.ready_for_query()) ->
+        new_data = %{data | caller: nil, waiting_cleanup: nil, pending_bin: nil}
+        {:next_state, :idle, new_data, {:reply, data.waiting_cleanup, :ok}}
+
+      byte_size(buffered_bin) > @cleanup_buffer_limit ->
```
mentels, line 407: any reason for the 64kb specifically? Just not keep the data on the process heap?

test/supavisor/db_handler_test.exs
```diff
@@ -590,4 +590,165 @@ defmodule Supavisor.DbHandlerTest do
...
+      data = %{
+        sock: {:gen_tcp, send},
+        mode: :session,
+        caller: self()
+      }
+
+      assert {:next_state, :waiting_cleanup, new_data, {:timeout, 5_000, :cleanup_timeout}} =
+               Db.handle_event({:call, {self(), make_ref()}}, :cleanup, :idle, data)
+
+      assert new_data.waiting_cleanup != nil
```
mentels:
```suggestion
      assert new_data.waiting_cleanup == pid
```

test/supavisor/db_handler_test.exs
```diff
@@ -590,4 +590,165 @@ defmodule Supavisor.DbHandlerTest do
...
+      data = %{
+        sock: {:gen_tcp, send},
+        mode: :session,
+        caller: self()
+      }
+
+      assert {:next_state, :waiting_cleanup, new_data, {:timeout, 5_000, :cleanup_timeout}} =
+               Db.handle_event({:call, {self(), make_ref()}}, :cleanup, :busy, data)
+
+      assert new_data.waiting_cleanup != nil
```
mentels: ditto

lib/supavisor/db_handler.ex
```diff
@@ -347,6 +360,24 @@ defmodule Supavisor.DbHandler do
     end
   end
 
+  def handle_event(:info, {proto, _, bin}, :waiting_cleanup, %{caller: caller} = data)
+      when is_pid(caller) and proto in @proto do
+    buffered_bin = data.pending_bin <> bin
+
+    cond do
+      String.ends_with?(buffered_bin, Server.ready_for_query()) ->
+        new_data = %{data | caller: nil, waiting_cleanup: nil, pending_bin: nil}
```
mentels, line 404: nitpick: we clean the `caller` but we don't set it

lib/supavisor/db_handler.ex
v0idpwn: Totally arbitrary amount. Needed to set a limit, went with 64KB. We have a process heap limit but (unless I'm wrong) it won't get off-heap binaries which can cause/have caused memory explosions.

test/supavisor/db_handler_test.exs
v0idpwn: Need to store the ref but will do that!

lib/supavisor/client_handler.ex
v0idpwn: >Can we rule out a scenario in wich an "unrelated" event arrives to the DbHandler while in :waiting_cleanup and cancels the DbHandler :cleanup_timeout?

Great point!!

I should've used `:state_timeout` instead of `:timeout` which ensures this does not happen. 

`:state_timeout` refers to the time spent in a state while `:timeout` refers to the time without processing messages (like in GenServer)

Will push a fix.

Other than that, I think we do a regular timeout in the gen:call just for sanity.

lib/supavisor/db_handler.ex
v0idpwn: Replied above, I'll just resolve this so we follow up in a single thread.

lib/supavisor/db_handler.ex
v0idpwn: `caller` is set at checkout time, I'm considering removing it part of the clean up

test/supavisor/db_handler_test.exs
v0idpwn: Done in https://github.com/supabase/supavisor/pull/837/commits/be9a89acea12b14e361125ae706187ba7ea82d5a

lib/supavisor/client_handler.ex
v0idpwn: Done in https://github.com/supabase/supavisor/pull/837/commits/4e39cba20885e8a3dd2ff3698acb535bab7e1860 and https://github.com/supabase/supavisor/pull/837/commits/0d252afd8a716d5c4bccb831d43ec3d5ad621601

v0idpwn (commented):
TODO: make sure we free the client before `terminate` so we reduce the time we impact the connection limit

lib/supavisor/client_handler.ex
```diff
@@ -798,6 +799,26 @@ defmodule Supavisor.ClientHandler do
     Logger.log(level, "ClientHandler: terminating with reason #{inspect(reason)}")
   end
 
+  defp maybe_cleanup_db_handler(state, data) do
+    if state == :idle and data.mode == :session and data.db_connection != nil and
+         !Supavisor.Helpers.no_warm_pool_user?(data.user) do
```
mentels, line 804: 🤯 takes a while to reason about this condition

lib/supavisor/client_handler.ex
```diff
@@ -798,6 +799,26 @@ defmodule Supavisor.ClientHandler do
...
+      Logger.debug("ClientHandler: Performing session cleanup before termination")
+      {pool, db_pid, _} = data.db_connection
+
+      # We unsubscribe to free up space for new clients during the cleanup time.
+      Supavisor.Manager.unsubscribe(data.id)
```
mentels, line 809: seems like the unsubscribe will only succeed if called on a node hosting the pool; is that guaranteed by the `data.mode == :session`?

lib/supavisor/client_handler.ex
mentels: Actually, there's yet another issue!

If the call fails - be it b/c of the `:timeout` or just b/c the `gen_statem` processes exists which is actually expected exec path here - the `gen_statem:call` exists the caller! And this will crash the `client_handler`.

So I guess to make it graceful we need to wrap it in the `try catch` OR make it async? (No, the latter sounds like a bad idea but "thinking out loud").

lib/supavisor/manager.ex
```diff
@@ -416,6 +446,28 @@ defmodule Supavisor.Manager do
...
+    :ets.foldl(
+      fn {ref, pid, _}, _ -> :ets.insert(pid_to_ref, {pid, ref}) end,
+      nil,
+      state.tid
+    )
+
+    new_state =
+      state
+      |> Map.put(:pid_to_ref, pid_to_ref)
+      |> Map.put_new(:waiting_for_secrets, [])
```
mentels: Why inserting the `:waiting_for_secrets`?

mentels (changes requested):
Changes looks good but as I pointed out in the comment, I _believe_ there's an issue.

lib/supavisor/client_handler.ex
v0idpwn: Yes, proxy clients will have `data.mode == :proxy`

lib/supavisor/manager.ex
v0idpwn: I was handling the whole code_change on this PR, there was something missing from another.

v0idpwn: I just removed them since we won't support soft deploying this.

mentels (approved):

PR #842, opened 2026-01-28
fix: invalidate cache on tenant creation

v0idpwn:
Previously, if a client tried to connect before a tenant was created, the `get_user_cache` function would cache the error. After the tenant was created, connection attempts would still fail until the cache TTL. Now the cache is cleared after inserts.

Additionally, cache and pool termination logic has been moved from the controller to the domain module.

### DIFF
--- a/docker-compose.db.yml
+++ b/docker-compose.db.yml
@@ -2,7 +2,7 @@ version: "3"
 
 services:
   db:
-    image: supabase/postgres:14.1.0.106
+    image: supabase/postgres:15.14.1.072
     container_name: supavisor-db
     ports:
       - "6432:5432"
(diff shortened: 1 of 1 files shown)


test/test_helper.exs
```diff
@@ -1,6 +1,11 @@
 {:ok, _} = Node.start(:"primary@127.0.0.1", :longnames)
 
-Cachex.start_link(name: Supavisor.Cache)
+# Cachex records the node name at startup for its internal routing. The
```
mentels, line 3: That's a surprise! 

I would put a little comment in the `application.ex` briefly mentioning this file - but it's definitely a neat pick.

BTW, great code clean-up - the `application.ex` got untangled from test specifics.

v0idpwn: I also got really surprised. It's due to Cachex distribution, even if we aren't using it distributed caches, we have this constraint about not being able to change the node name after it was started.

PR #846, opened 2026-01-29
Add HTTP API to list tenants' bans

mentels:
This PR adds an HTTP API `GET     /api/tenants/:external_id/bans` that lists bans for a tenant.

### DIFF
--- a/lib/supavisor_web/open_api_schemas.ex
+++ b/lib/supavisor_web/open_api_schemas.ex
@@ -323,7 +323,7 @@ defmodule SupavisorWeb.OpenApiSchemas do
       required: [:operation, :blocked_until],
       example: %{
         operation: "auth_error",
-        blocked_until: 1706549400
+        blocked_until: 1_706_549_400
       }
     })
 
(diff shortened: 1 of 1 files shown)


lib/supavisor/circuit_breaker.ex
```diff
@@ -100,6 +100,22 @@ defmodule Supavisor.CircuitBreaker do
...
+  Returns an empty list if no operations are blocked for the key or the key does not exist.
+  """
+  @spec blocked(term()) :: [{atom(), integer()}]
+  def blocked(key) do
+    now = System.system_time(:second)
+
+    :ets.select(@table, [
+      {{{key, :"$1"}, %{blocked_until: :"$2"}},
+       [{:andalso, {:is_integer, :"$2"}, {:>, :"$2", now}}], [{{:"$1", :"$2"}}]}
+    ])
```
v0idpwn, line 116: I don't think this will work for `auth_error` bans, which is the main (only?) operation we are interested in. Their key is not `tenant` but `{tenant, ip}`.

PR #849, opened 2026-02-05
chore: test upgrade make target

v0idpwn:
Adds a make target that allows testing the upgrade from a version to the current HEAD.


mentels (approved):
Really cool!

PR #850, opened 2026-02-05
fix(tenant_supervisor): set child_spec type to :supervisor

v0idpwn:
The child_spec was missing `type: :supervisor`, causing the release handler to treat TenantSupervisor as a worker and skip its children during hot upgrades.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,9 @@ pgbench:
 clean:
 	rm -rf _build && rm -rf deps
 
+test_upgrade:
+	scripts/test_upgrade.sh $(FROM)
+
 dev_release:
 	mix deps.get && mix compile && mix release supavisor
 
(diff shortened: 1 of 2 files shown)


test/supavisor/tenant_supervisor_test.exs
```diff
@@ -0,0 +1,14 @@
...
+
+  test "tenant supervisor child spec" do
+    child_spec = Supavisor.TenantSupervisor.child_spec(%{id: nil})
+
+    # Type being `:supervisor` is important for hot code reloading
+    # to affect it's children
+    assert child_spec.type == :supervisor
+
+    # Restart being `:transient` avoids a bad pool impacting others
+    assert child_spec.restart == :transient
```
mentels, line 12: From what I see it didn't change, it is `:transient`.

v0idpwn: Oh yeah, I just took the opportunity to cover it since it's not something we test in any other way AFAIK

PR #851, opened 2026-02-07
feat(network-bans): Add HTTP API for network bans mgmt; make auth bans cluster wide

mentels:
Add `{GET, DELETE} /api/tenants/:external_id/network_bans` for network bans management.

Modify the `CiricuitBreaker.{record_failure/2,clear/2}` to propagate bans / clearing bans across the cluster with `:erpc.call`.

Supersedes #848 and #846

### DIFF
--- a/lib/supavisor/circuit_breaker.ex
+++ b/lib/supavisor/circuit_breaker.ex
@@ -43,9 +43,6 @@ defmodule Supavisor.CircuitBreaker do
 
   @doc """
   Records a failure for a given key and operation.
-
-  For the :auth_error, it the circuit is opened, it is opened across all the nodes
-  in the cluster.
   """
   @spec record_failure(term(), atom()) :: :ok
   def record_failure(key, operation) when is_atom(operation) do
--- a/lib/supavisor/tenants.ex
+++ b/lib/supavisor/tenants.ex
@@ -719,37 +719,8 @@ defmodule Supavisor.Tenants do
     ClusterTenants.changeset(cluster_tenants, attrs)
   end
 
-  # TODO: remove?
   @doc """
-  Returns a list of currently blocked operations for a given tenant external_id.
-
-  Each entry is a map with :operation and :blocked_until keys. If there are
-  no blocked operations for a tenant, an empty list is returned.
-
-  ## Examples
-
-      iex> list_bans("tenant_external_id")
-      [%{operation: :some_operation, blocked_until: 1620000000}, ...]
-  """
-  @spec list_bans(String.t()) ::
-          {:ok,
-           [
-             %{required(:operation) => atom(), required(:blocked_until) => integer()}
-           ]}
-          | {:error, :tenant_not_found}
-  def list_bans(external_id) when is_binary(external_id) do
-    if get_tenant_by_external_id(external_id) do
-      list_network_bans(external_id)
-    else
-      {:error, :tenant_not_found}
-    end
-  end
-
-  @doc """
-  Returns a list of network bans (auth_error) for a given tenant.
-
-  This specifically lists IP-based authentication error bans across the cluster
-  along with their unblock unix timestamp.
+  Returns a list of network bans for a given tenant.
 
   ## Examples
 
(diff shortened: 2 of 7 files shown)


lib/supavisor/circuit_breaker.ex
```diff
@@ -100,15 +105,119 @@ defmodule Supavisor.CircuitBreaker do
...
+    :ok
+  end
+
+  # Clears circuit breaker state globally for :auth_error operation.
+  # Propagates the clear operation to all nodes in the cluster.
+  defp maybe_clear_globally(key, :auth_error = operation) do
+    nodes = Node.list()
+
+    nodes
+    |> :erpc.multicall(__MODULE__, :clear, [key, operation], 10_000)
```
v0idpwn: This will recurse indefinitely.

clear/2 calls maybe_clear_globally/2, which calls clear/2 on the remote nodes, which  will call maybe_clear_globally/2` again.

lib/supavisor/circuit_breaker.ex
```diff
@@ -100,15 +105,119 @@ defmodule Supavisor.CircuitBreaker do
...
+    ])
+    |> Enum.map(fn {{key, _operation}, %{blocked_until: blocked_until}} ->
+      {key, blocked_until}
+    end)
+  end
+
   @doc """
   Clears circuit breaker state for a key and operation.
+
+  For :auth_error operation clears the state on all the cluster nodes.
```
v0idpwn: I think it would be nicer to add a new config in the CB config map for this instead of hardcoding behaviour for `auth_error`. 

Then you can do `@thresholds[:propagate_bans?]` and use it to decide whether to propagate (maybe rename the `@thresholds` attr too)

lib/supavisor/tenants.ex
```diff
@@ -717,4 +718,80 @@ defmodule Supavisor.Tenants do
...
+             %{required(:operation) => atom(), required(:blocked_until) => integer()}
+           ]}
+          | {:error, :tenant_not_found}
+  def list_bans(external_id) when is_binary(external_id) do
+    if get_tenant_by_external_id(external_id) do
+      list_network_bans(external_id)
+    else
+      {:error, :tenant_not_found}
+    end
+  end
```
v0idpwn: Let's remove this?

lib/supavisor_web/open_api_schemas.ex
```diff
@@ -288,4 +288,127 @@ defmodule SupavisorWeb.OpenApiSchemas do
...
+      },
+      required: [:operation, :blocked_until],
+      example: %{
+        operation: "auth_error",
+        blocked_until: 1_706_549_400
+      }
+    })
+
+    def response, do: {"Ban Response", "application/json", __MODULE__}
+  end
```
v0idpwn: I think this (and `CircuitBreakerOperation`) are dead code?

lib/supavisor/circuit_breaker.ex
```diff
@@ -100,15 +105,119 @@ defmodule Supavisor.CircuitBreaker do
...
   end
 
+  # Propagates the :auth_error ban to all nodes except for the current node.
+  # Does NOT run on the caller's node.
+  defp maybe_open_globally(key, :auth_error = operation, blocked_until)
+       when is_integer(blocked_until) do
+    nodes = Node.list()
+
+    nodes
+    |> :erpc.multicall(__MODULE__, :open, [key, operation, blocked_until], 10_000)
```
v0idpwn: Should we maybe be using `multicast` here? WDYT?

lib/supavisor/circuit_breaker.ex
mentels: Great catch!

lib/supavisor/circuit_breaker.ex
mentels: I was considering the `multicast` but decided not to use it b/c 
>multicast() returns immediately after the cast requests have been sent. Any failures beside bad arguments are silently ignored.
- so we'd loose any trace of potential errors.

PR #852, opened 2026-02-09
test:add test for tenant in different az (coverage)

v0idpwn:
This reproduces an error with our logic where we can't send the secrets to another node if the version changed, because they are wrapped in an anonymous function that can't be unwrapped in the other side.


mentels (approved):
The changes are OK but the PR description doesn't match IMO :)

v0idpwn:
Need #853

PR #859, opened 2026-02-11
WIP

mentels:
This moves the `:erlang.binary_to_term(binary)` out of the async process to the connection processes hoping that not-copying the data but just the refC bin pointer is more efficient.

However laptop-scale tests haven't revealed any significant improvement.


v0idpwn:
Closed in favor of #875

PR #891, opened 2026-03-23
feat: add app_name in active connections metric, streamline app_name

v0idpwn:
Adds the application name provided in the connection startup to the ClientHandler metadata, and emit it in telemetry events, conditional to a feature flag being enabled.

Ensures ClientHandler always have the user-provided application_name or "", in proxy and non-proxy scenarios.

Previously, the ClientHandler would default to the "Supavisor" app name when one was not provided.  Ensures that the postgres-side backend always has "Supavisor" app name for transaction mode, regardless of what's the app_name supavisor-side. For session mode, use a mix of the Supavisor app name to the client provided app name.


mentels (approved):

lib/supavisor/protocol/server.ex
```diff
@@ -207,8 +207,40 @@ defmodule Supavisor.Protocol.Server do
...
+    ])
+  end
+
   # Internal functions
 
+  defp encode_bind(params) do
+    encoded_params =
+      Enum.map(params, fn value ->
+        value = IO.iodata_to_binary(value)
+        [<<byte_size(value)::32>>, value]
```
v0idpwn:
```suggestion
        [<<IO.iodata_length(value)::32>>, value]
```

test/supavisor/monitoring/tenant_test.exs
```diff
@@ -153,12 +158,100 @@ defmodule Supavisor.PromEx.Plugins.TenantTest do
...
+            {^ref, {[:supavisor, :connections], measurement, meta}} ->
+              {:cont, [{measurement, meta} | acc]}
+          after
+            100 -> {:halt, acc}
+          end
+        end)
+
+      assert length(events) == 3
+
+      by_app = Map.new(events, fn {%{active: count}, %{app_name: app}} -> {app, count} end)
```
Snehil-Shah:
```suggestion
      by_app = Map.new(events, fn {%{active: count}, %{app_name: app}} -> {app, count} end)

      assert map_size(by_app) == 3
```

@v0idpwn In our last test run, the background metrics poller happened to slip into the 100ms window. I think the above can be a better way to assert, as we will just be checking if the expected groups came in or not? WDYT?

Snehil-Shah: Also, I think we can follow the familiar pattern established in #1026 in this test case as well, to prevent race?

v0idpwn: Addressed on https://github.com/supabase/supavisor/pull/891/changes/f4fa21eaee73468737b5775bcd856821d25d5bca thanks!

PR #899, opened 2026-03-30
fix: SNI-based tenant routing 

arcinston:
SNI-based tenant routing silently fails across all tested versions (2.7.4, 2.8.8, 2.9.0-rc). When a client connects with a plain username (e.g. user=appuser) and relies on the TLS SNI hostname (e.g. demo-db.local.arc.cloud) for tenant resolution, try_get_sni() returns nil and the connection either falls back to dot-delimited username routing or crashes with "comparison with nil is forbidden" in Ecto.Query.Builder.

## What is the new behavior?
SNI-based tenant routing works end-to-end. Clients can connect with a plain username and the tenant is correctly resolved from the TLS SNI hostname. Pool identity and config lookups use the resolved tenant's external_id.
## Additional context
Two separate bugs in lib/supavisor/client_handler.ex:

The :ssl.handshake call doesn't include an sni_fun callback. Erlang OTP's :ssl module requires sni_fun to be registered during the handshake for the SNI hostname to be captured and returned by :ssl.connection_information/2. Without it, the handshake succeeds but sni_hostname is never stored. Fixed by adding sni_fun: fn _hostname -> [certs_keys: certs_keys] end to the handshake opts.

After SNI extraction works, the pool identity struct is still built with tenant: tenant_or_alias, where tenant_or_alias is parsed from the username and is nil for SNI-only connections. This nil propagates through start_local_pool → get_pool_config_cache → get_pool_config, where Ecto rejects the nil comparison. Fixed by falling back to info.tenant.external_id when tenant_or_alias is nil.

## Personal Context
I was trying to run Supavisor with sni routing on K8s but it kept failing for all tested versions (2.7.4, 2.8.8, 2.9.0-rc-3). Tried looking into the code & fixing this, this change did fix the error 

[logfile.txt](https://github.com/user-attachments/files/26358228/logfile.txt)
adding the error log i recieved when trying to connect using SNI routing - it would fail just after the correct password was accepted and just before the handshake

Hoping this helps!

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -158,7 +158,8 @@ defmodule Supavisor.ClientHandler do
 
       opts = [
         verify: :verify_none,
-        certs_keys: certs_keys
+        certs_keys: certs_keys,
+        sni_fun: fn _hostname -> [certs_keys: certs_keys] end
       ]
 
       case :ssl.handshake(elem(sock, 1), opts) do
@@ -228,10 +229,12 @@ defmodule Supavisor.ClientHandler do
       {:ok, info} ->
         upstream_tls = upstream_tls(info.tenant, effective_ssl)
 
+        resolved_tenant = tenant_or_alias || info.tenant.external_id
+
         id =
           Supavisor.id(
             type: type,
-            tenant: tenant_or_alias,
+            tenant: resolved_tenant,
             user: user,
             mode: data.mode,
             db: db_name,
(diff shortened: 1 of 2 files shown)


v0idpwn (commented):
Thanks for the PR! The code looks good to me, but can we cover it with a test?

arcinston:
@v0idpwn hi hi - added tests!
had fun setting up the whole repository and running tests!

arcinston:
hi hi @v0idpwn 
was curious if supabase directly use this version of supavisor or is the oss version different from the production version?
cause SNI not working would be a major regression?

v0idpwn:
@arcinston the code looks great, will merge and ship! 

We are in v2.8 in production in most of our clusters, but we direct to connect via `{username}.{tenant}`in our interface/docs, hence the SNI hostname regression went unnoticed. I believe none or close to none users use this feature in our platform.

v0idpwn (approved):

PR #902, opened 2026-04-01
chore: hard deploy v2.9.0-rc.4 on usw21

v0idpwn:


mentels (approved):

deploy/upgrade/prod.config
```diff
@@ -1,17 +1,13 @@
 [
     {upgrade_from, "2.8.6"},
     {upgrade_previous, "2.8.6"},
```
deepthi, line 3: Not sure if this affects anything, but do these lines need to change to 2.8.8?

PR #903, opened 2026-04-02
fix: return NoTenantIdentifierError when external_id is nil and sni host doesn't match

v0idpwn:
When external_id is nil and sni hostname doesn't match, the error formats as {user}.{sni_hostname} which is meaningless to the user. Return NoTenantIdentifierError instead.


v0idpwn:
Fyi @arcinston, I think the message became a bit confusing with the SNI hostname fallback.

You'd get messages such as `tenant/user postgres.aws-1-us-west-2.pooler.supabase.com not found`, which is not right, the sni hostname is not the tenant name. I think this that I'm pushing is a bit better, but its not perfect for the sni-based routing. 

If you think we can have a middleground that gives good messages for both, we are open for PRs :)

arcinston:
agreed , this makes more sense

mentels (approved):

Issue #172, opened 2023-10-02
Better checkout timeout error message


chasers:
Inspired by the Elixir `DBConnection.ConnectionError`

```
Pool checkout timeout error. Request was dropped after #{n}ms. This means requests are coming in and your connection pool cannot serve them fast enough. You can address this by:

  1. Ensuring your database is available and that you can connect to it
  2. Tracking down slow queries and making sure they are running fast enough
  3. Increasing the pool_size (although this increases resource consumption)
  4. Upgrading your database

Review the Supabase performance tuning guide for Postgres: https://supabase.com/docs/guides/platform/performance
```

v0idpwn:
We added error codes instead.

PR #904, opened 2026-04-03
fix(errors): improve checkout timeout error with actionable guidance

SriSomanaath:
## What does this PR do?

This addresses #172 by making the checkout timeout error message actually helpful. Right now when users hit a pool timeout, they just get a terse message that doesn't tell them what to do about it. I've updated it to follow the same pattern that DBConnection uses - explaining what's happening and giving concrete steps to fix it.

## The problem

Current error message:
```
(ECHECKOUTTIMEOUT) unable to check out connection from the pool after 5000ms in Transaction mode
```

Users see this and have no idea what to do next. They end up searching forums, reading docs, or just increasing timeouts randomly hoping it helps.

## The solution

New error message:
```
(ECHECKOUTTIMEOUT) connection not available and checkout timed out after 5000ms in Transaction mode. This means requests are coming in and your connection pool cannot serve them fast enough.

You can address this by:

  1. Ensuring your database is available and that you can connect to it
  2. Tracking down slow queries and making sure they are running fast enough
  3. Increasing the pool_size (note: this increases database resource consumption, but Transaction mode allows connection reuse between queries)
  4. Increasing the checkout timeout (note: this increases latency under load)

See https://supabase.com/docs/guides/database/connecting-to-postgres for more information on connection pooling.
```

The hints are also mode-specific - Session mode users get reminded that each client holds a dedicated connection, so they understand why pool_size matters differently for them.

## Other changes

- Added a separate `log_message/1` that keeps logs concise (nobody wants multi-line error messages in their log aggregator)
- Changed log level to `:warning` instead of `:error` since timeouts are usually transient and self-resolving
- Added tests covering all the new functionality

## Testing

Added a new test file at `test/supavisor/errors/checkout_timeout_error_test.exs` that covers:
- Error message content for both modes
- All 4 remediation steps being present
- Mode-specific pool_size hints
- Log message format
- Log level
- Postgres error structure

## Notes

I noticed PR #896 attempted something similar but was closed. I took a slightly different approach here - keeping the log messages short while making the client-facing error detailed. Happy to adjust based on feedback.

### DIFF
--- a/lib/supavisor/errors/checkout_timeout_error.ex
+++ b/lib/supavisor/errors/checkout_timeout_error.ex
@@ -1,8 +1,12 @@
 defmodule Supavisor.Errors.CheckoutTimeoutError do
   @moduledoc """
   This error is returned when unable to check out a connection from the pool due to timeout.
-  This happens when all connections in the pool are busy and none become available within the checkout timeout period.
-  Applies to both session and transaction modes.
+
+  This happens when all connections in the pool are busy and none become available within
+  the checkout timeout period. The error provides actionable guidance to help users
+  diagnose and resolve connection pool exhaustion issues.
+
+  Applies to both session and transaction modes, with mode-specific remediation hints.
   """
 
   use Supavisor.Error, [:mode, :timeout_ms, code: "ECHECKOUTTIMEOUT"]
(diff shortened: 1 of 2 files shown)


v0idpwn:
Thanks @SriSomanaath but instead of the long error message we added an error code on the latest release. We will be soon releasing a public documentation with error codes and action items which we think is a better way to handle this.

I will be closing the issue as well.

PR #905, opened 2026-04-07
perf: optimizations around auth

v0idpwn:
(1) If connecting over TLS, use password instead of scram-sha-256. One less round trip and less expensive for the client. Using scram-sha-256 is redundant with the TLS cryptography. Previously, this was conditional to use_jit tenants.

(2) If Supavisor has plaintext password and the database asks for scram-sha-256, cache the scram-sha-256 secret so we don't need to recompute PBKDF2.


mentels (approved):
LGTM

mentels (approved):

PR #907, opened 2026-04-08
feat(tenants): terminate active connections when tenant is banned

mentels:
## Summary

 * Respect the `banned_until`
 * Immediately drops all active client connections with `EBANNED` error when a tenant is banned.

Also adds a test demonstrating that a banned tenant, after deletion, is re-created unbanned.


lib/supavisor.ex
```diff
@@ -160,6 +160,43 @@ defmodule Supavisor do
     :erpc.multicall([node() | Node.list()], Supavisor, :dirty_terminate, [tenant], 60_000)
   end
 
+  @doc """
+  Terminates all client handlers for a given tenant locally.
+
+  Uses the id record's field index to efficiently match clients for the given tenant.
+  """
+  def terminate_client_handlers(tenant) do
```
v0idpwn: If we shutdown the pool supervision tree, all client handlers, which monitor the Manager, should  stop with `manager_shutdown`, no? Why do we need to shut them down in a separate step?

mentels: Great point! Haven't thought about approaching it that way; will update the PR.

v0idpwn: I think depending on how we finish it, we can even close them with an error via graceful shutdown? Then the existing connections could get EBANNED

lib/supavisor.ex
```diff
@@ -133,31 +134,66 @@ defmodule Supavisor do
   end
 
   @doc """
-  During netsplits, or due to certain internal conflicts, :syn may store inconsistent data across the cluster.
-  This function terminates all connection trees related to a specific tenant.
+  Terminate all connection trees for a tenant across the cluster
+
+  During netsplits, or due to certain internal conflicts, :syn may store inconsistent
+  data across the cluster.
```
mentels, line 140: This comment is a little bit mysterious but left it just in case :)

lib/supavisor.ex
```diff
@@ -133,31 +134,66 @@ defmodule Supavisor do
...
+
+  If `error` is provided, `Manager.shutdown_with_error/2` will be used to first
+  terminate the clients with the provided error, before terminating a pool -
+  which is done asynchronously from the `Manager`.
+
+  There's a race, in which the pool may be terminated before the clients terminate with the error
+  - then their connection will be closed forcibly without the error message.
```
mentels, line 147: I thought this is fine :this-is-fine:

lib/supavisor.ex
```diff
@@ -133,31 +134,66 @@ defmodule Supavisor do
...
+
+  def dirty_terminate(tenant, error, _timeout) do
+    Supavisor.Registry.TenantSups
+    |> Registry.lookup(tenant)
+    |> Enum.reduce(%{}, fn {_pid, id(user: user) = id}, acc ->
       resp = %{
-        stop: stop,
+        stop:
+          if pid = get_local_manager(id) do
+            Manager.shutdown_with_error(pid, TenantBannedError.postgres_error(error))
```
mentels, line 178: The `Manager` does the `Supervisor.stop/1` which is equivalent to  `Supervisor.stop(pid, :shutdown, :infinity)`

v0idpwn (approved):

PR #916, opened 2026-04-13
fix: drop auth log

v0idpwn:
We don't need it anymore.

This PR also:

- Removes `reint_funs` from upgrade instructions (would break soft deployments)
- Removes `+JPperf true` from vm args when not on linux (would break release startup on non-Linux platforms)
- Ensures that if upgrading from v2.9.0 to v2.9.1 the handler is dropped


lib/supavisor/hot_upgrade.ex
```diff
@@ -27,20 +27,31 @@ defmodule Supavisor.HotUpgrade do
 
   @spec up(app(), version_str(), version_str(), [appup()], any()) :: [appup()]
   def up(_app, _from_vsn, to_vsn, appup, _transform) do
+    appup =
+      Enum.reject(appup, fn
+        {:load_module, __MODULE__} -> true
+        {:load_module, __MODULE__, _} -> true
```
mentels, line 33: What does it change?

I can see you load the current module below.

mentels (approved):
LGTM, although I think it would be worth documenting somewhere how to restore the `access.log` functionality in case it turns out useful for us in the future. 

Eg if we want to ban tenant-related IPs at the lower level.

lib/supavisor/hot_upgrade.ex
v0idpwn: We now load this module before running the function instead of after

v0idpwn: I think if we want to reintroduce this functionality we should probably go a different path, e.g.: making a plugin to interact with fail2ban APIs and connect it to the circuit breaker.

PR #921, opened 2026-04-15
fix(compatibility): accept query encoded options

v0idpwn:
These aren't accepted by postgres, but due to historical reasons, must be supported by Supavisor

This commit falls back to try them when there are no options found in the regular parsing path


lib/supavisor/protocol/client.ex
```diff
@@ -29,8 +29,17 @@ defmodule Supavisor.Protocol.Client do
         fields
         |> Enum.chunk_every(2)
         |> Enum.map(fn
-          ["options" = k, v] -> {k, Supavisor.Protocol.StartupOptions.parse(v)}
-          [k, v] -> {k, v}
+          ["options" = k, v] ->
+            parsed = Supavisor.Protocol.StartupOptions.parse(v)
+            # Compatibility: options are not following the standard format, try URL decoding them
```
v0idpwn:
```suggestion
            # Compatibility: options are not following the standard format, try query_decoding them
```

mentels (approved):

test/integration/proxy_test.exs
```diff
@@ -938,6 +938,36 @@ defmodule Supavisor.Integration.ProxyTest do
...
+      username: db_conf[:username]
+    ]
+
+    assert {:ok, proxy} =
+             Postgrex.start_link(base_opts ++ [parameters: [options: "--reference=#{tenant}"]])
+
+    assert %Postgrex.Result{rows: [[1]]} = Postgrex.query!(proxy, "SELECT 1", [])
+
+    assert {:ok, proxy2} =
+             Postgrex.start_link(base_opts ++ [parameters: [options: "reference=#{tenant}"]])
```
ferhatelmas, line 959: I think it's better to add a test that mixes them `reference=tenant --search_path=public`

v0idpwn: That won't work :( 
Much like `reference=tenant -c search_path=public`. But I'm hoping that this scenario is harder to run in.

lib/supavisor/protocol/client.ex
```diff
@@ -29,8 +29,17 @@ defmodule Supavisor.Protocol.Client do
...
         |> Enum.map(fn
-          ["options" = k, v] -> {k, Supavisor.Protocol.StartupOptions.parse(v)}
-          [k, v] -> {k, v}
+          ["options" = k, v] ->
+            parsed = Supavisor.Protocol.StartupOptions.parse(v)
+            # Compatibility: options are not following the standard format, try query_decoding them
+            if map_size(parsed) > 0 do
+              {k, parsed}
+            else
+              {k, URI.decode_query(v)}
```
ferhatelmas: afaik, storage doesn't need but [:rfc3986](https://hexdocs.pm/elixir/URI.html#decode_query/3) encoding for completeness

lib/supavisor/protocol/client.ex
```diff
@@ -29,8 +29,18 @@ defmodule Supavisor.Protocol.Client do
...
-          ["options" = k, v] -> {k, Supavisor.Protocol.StartupOptions.parse(v)}
-          [k, v] -> {k, v}
+          ["options" = k, v] ->
+            parsed = Supavisor.Protocol.StartupOptions.parse(v)
+
+            # Compatibility: options are not following the standard format, try query_decoding them
+            if map_size(parsed) > 0 do
+              {k, parsed}
+            else
+              {k, URI.decode_query(v)}
```
v0idpwn:
```suggestion
              {k, URI.decode_query(v, :rfc3986)}
```

lib/supavisor/protocol/client.ex
```diff
@@ -29,8 +29,18 @@ defmodule Supavisor.Protocol.Client do
...
-          ["options" = k, v] -> {k, Supavisor.Protocol.StartupOptions.parse(v)}
-          [k, v] -> {k, v}
+          ["options" = k, v] ->
+            parsed = Supavisor.Protocol.StartupOptions.parse(v)
+
+            # Compatibility: options are not following the standard format, try query_decoding them
+            if map_size(parsed) > 0 do
+              {k, parsed}
+            else
+              {k, URI.decode_query(v, :rfc3986)}
```
v0idpwn:
```suggestion
              {k, URI.decode_query(v, %{}, :rfc3986)}
```

ferhatelmas: since we updated this, shall we add a test with tenant including `+`?

test/integration/proxy_test.exs
ferhatelmas: I agree it's less likely. It would be the best if we could prevent but less a concern

PR #952, opened 2026-04-29
chore: upgrade to v2.9.2 in ap-southeast-2-sec

mentels:
platform: https://github.com/supabase/platform/pull/32227


mentels:
the cluster is already up to date

PR #962, opened 2026-05-04
fix: add timeout on connection, better error for timeout on auth query

v0idpwn:
Also sets max proxy connections to infinity.

### DIFF
--- a/lib/supavisor/application.ex
+++ b/lib/supavisor/application.ex
@@ -74,7 +74,11 @@ defmodule Supavisor.Application do
           key,
           :ranch_tcp,
           %{
-            max_connections: String.to_integer(System.get_env("MAX_CONNECTIONS") || "75000"),
+            max_connections:
+              case System.get_env("MAX_CONNECTIONS") do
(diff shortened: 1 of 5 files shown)


lib/supavisor/application.ex
```diff
@@ -74,7 +74,11 @@ defmodule Supavisor.Application do
           key,
           :ranch_tcp,
           %{
-            max_connections: String.to_integer(System.get_env("MAX_CONNECTIONS") || "75000"),
+            max_connections:
+              case System.get_env("MAX_CONNECTIONS") do
+                nil -> :infinity
```
deepthi, line 79: What is the rationale for this?

v0idpwn: I should've split this into a separate pr 🙈 
This artificially limits the number of connections we accept and it doesn't give visibility on the non-accepted sockets. We never observed it in production (that we know of), but while investigating the incident I noticed it and am preemptively increasing it.

deepthi (approved):

PR #965, opened 2026-05-04
chore: temp-revert optimization to use password instead of sasl on tls

v0idpwn:


deepthi:
Which prior PR is this reverting, if any? That is useful context for review.

v0idpwn:
Not reverting the whole PR, just the auth method selection on #905

deepthi (approved):

PR #969, opened 2026-05-06
feat: add metrics for auth_query and SecretChecker erpc calls

mentels:
## Summary

Closes POOLER-392.

Adds four distribution metrics to improve observability of `auth_query` and `SecretChecker` operations:

- **`supavisor_auth_query_connection_duration`** — Postgrex connection setup time, emitted from `AuthQuery.start_link/2`
- **`supavisor_auth_query_query_duration`** — queue wait + query RTT, emitted from `AuthQuery.fetch_user_secret/3`; logs a warning when execution exceeds 1s
- **`supavisor_secret_checker_get_secrets_duration_{local,remote}`** — full erpc round-trip for `get_secrets`, split by locality (same pattern as `pool_checkout`)
- **`supavisor_secret_checker_update_credentials_duration_{local,remote}`** — same for `update_credentials`

All metrics tagged with `status` only — no per-tenant tags to avoid cardinality explosion.

Also fixes `Logger.error` in `do_get_secrets/1` to include `project: tenant` structured metadata (the function runs in the erpc caller process, not the SecretChecker GenServer, so the metadata was previously missing).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/supavisor/auth_query.ex
+++ b/lib/supavisor/auth_query.ex
@@ -31,7 +31,8 @@ defmodule Supavisor.AuthQuery do
       socket_options: [ip_version],
       queue_target: 1_000,
       queue_interval: 5_000,
-      ssl_opts: ssl_opts
+      ssl_opts: ssl_opts,
+      connection_listeners: {[Supavisor.ConnectionListener], start}
     ]
     |> Postgrex.start_link()
     |> case do
@@ -39,12 +40,13 @@ defmodule Supavisor.AuthQuery do
         {:ok, pid}
 
       :ignore ->
+        telemetry_connection_stop(start, :connection_failed)
         {:error, %AuthQueryError{reason: :connection_failed, details: "connection ignored"}}
 
       {:error, reason} ->
+        telemetry_connection_stop(start, :connection_failed)
         {:error, %AuthQueryError{reason: :connection_failed, details: inspect(reason)}}
     end
-    |> tap(&telemetry_connection_stop(start, &1))
   end
 
   @doc """
@@ -181,9 +183,8 @@ defmodule Supavisor.AuthQuery do
     [verify: :verify_none]
   end
 
-  defp telemetry_connection_stop(start, result),
-    do:
-      Telem.auth_query_connection_stop(System.monotonic_time() - start, telemetry_status(result))
+  defp telemetry_connection_stop(start, error),
+    do: Telem.auth_query_connection_stop(System.monotonic_time() - start, telemetry_status(error))
 
   defp telemetry_query_stop(start, result),
     do: Telem.auth_query_query_stop(System.monotonic_time() - start, telemetry_status(result))
--- a/lib/supavisor/auth_query.ex
+++ b/lib/supavisor/auth_query.ex
@@ -20,32 +20,31 @@ defmodule Supavisor.AuthQuery do
     ip_version = Helpers.ip_version(tenant.ip_version, tenant.db_host)
     start = System.monotonic_time()
 
-    result =
-      case Postgrex.start_link(
-             hostname: tenant.db_host,
-             port: tenant.db_port,
-             database: tenant.db_database,
-             password: manager.db_password,
-             username: manager.db_user,
-             parameters: [application_name: "Supavisor (auth_query)"],
-             ssl: tenant.upstream_ssl,
-             socket_options: [ip_version],
-             queue_target: 1_000,
-             queue_interval: 5_000,
-             ssl_opts: ssl_opts
-           ) do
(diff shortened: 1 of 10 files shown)


lib/supavisor/monitoring/telem.ex
```diff
@@ -5,6 +5,7 @@ defmodule Supavisor.Monitoring.Telem do
   require Supavisor
 
   @disabled Application.compile_env(:supavisor, :metrics_disabled, false)
+  @slow_auth_query_ms 1_000
```
mentels, line 8: I'm not sure if this is the right number, we can see what the times are in the first cluster we deploy to.

lib/supavisor/monitoring/tenant.ex
```diff
@@ -191,6 +192,68 @@ defmodule Supavisor.PromEx.Plugins.Tenant do
     )
   end
 
+  defp client_auth_metrics do
```
mentels, line 195: technically, these are not per-tenant metrics but thought it's not worth creating a new module just for them

test/supavisor/auth_query_test.exs
```diff
@@ -129,8 +130,10 @@ defmodule Supavisor.AuthQueryTest do
       {:ok, conn} =
         Postgrex.start_link(
           hostname: "invalid.nonexistent.host",
-          port: 5432,
-          database: "test"
+          port: 5433,
+          database: "test",
+          queue_target: 1,
+          queue_interval: 4
```
mentels, line 136: making the test fast

lib/supavisor/secret_checker.ex
```diff
@@ -40,7 +41,10 @@ defmodule Supavisor.SecretChecker do
             {:error, %AuthQueryError{reason: :timeout}}
 
           :exit, reason ->
-            Logger.error("SecretChecker: get_secrets call exited: #{inspect(reason)}")
+            Logger.error("SecretChecker: get_secrets call exited: #{inspect(reason)}",
+              project: Supavisor.id(id, :tenant)
+            )
```
mentels, line 46: we were missing the project md if that was an RPC call

lib/supavisor/auth_query.ex
```diff
@@ -17,20 +18,23 @@ defmodule Supavisor.AuthQuery do
...
+      password: manager.db_password,
+      username: manager.db_user,
+      parameters: [application_name: "Supavisor (auth_query)"],
+      ssl: tenant.upstream_ssl,
+      socket_options: [ip_version],
+      queue_target: 1_000,
+      queue_interval: 5_000,
+      ssl_opts: ssl_opts
+    ]
+    |> Postgrex.start_link()
```
v0idpwn, line 37: Connection in postgrex is async unless we pass the sync flag, this function returns a pid even without the connection being complete, so the measurement won't get slow connections

lib/supavisor/monitoring/telem.ex
```diff
@@ -117,6 +118,50 @@ defmodule Supavisor.Monitoring.Telem do
...
+    )
+  end
+
+  @spec auth_query_query_stop(integer(), :ok | :error) :: :ok
+  def auth_query_query_stop(duration, status) do
+    duration_ms = System.convert_time_unit(duration, :native, :millisecond)
+
+    if duration_ms > @slow_auth_query_ms do
+      Logger.warning("auth_query took over #{@slow_auth_query_ms}ms (#{duration_ms}ms)")
+    end
```
v0idpwn, line 136: Really like this!

test/supavisor/auth_query_test.exs
```diff
@@ -2,6 +2,7 @@ defmodule Supavisor.AuthQueryTest do
   use ExUnit.Case, async: true
```
v0idpwn, line 2: I think after the changes in this file, this test case must be sync?

lib/supavisor/secret_checker.ex
```diff
@@ -161,9 +165,12 @@ defmodule Supavisor.SecretChecker do
   defp jitter, do: :rand.uniform(div(@interval, 10))
 
   defp erpc_call_node(id, mod, fun, args) do
+    start = System.monotonic_time()
+
     case Supavisor.get_global_sup(id) do
       nil ->
         {:error, :not_started}
+        |> tap(&telemetry_stop(fun, start, &1, :local))
```
v0idpwn, line 173: I think `not_started` as in "pool not up" should not be tagged as error in telemetry but as something else... Maybe should review the return values of the function too and do telemetry via reason, e.g.: `not_started`, `exited`, `timeout`, ...?

PR #975, opened 2026-05-07
chore: rollback cac11 to v2.9.2

v0idpwn:
Safety PR to revert #971


deepthi (approved):

deploy/upgrade/prod.config
```diff
@@ -1,7 +1,7 @@
 [
     {upgrade_from, "2.9.0"},
```
deepthi: Does this need to change?

v0idpwn: nope, ignored on hard (rollback is going to be hard if it happens)

PR #978, opened 2026-05-12
chore(deps): bump phoenix to 1.7.23

MorganaFuture:
## What kind of change?
Security dependency bump.

## Current behavior
Phoenix 1.7.20 is affected by GHSA-628h-q48j-jr6q (high):
long-poll NDJSON body splitting causes large memory allocation. 
`mix deps.audit` fails CI.

## New behavior
Bump Phoenix to 1.7.23 (first patched: 1.7.22). `~> 1.7.2` in
mix.exs already allows this — only mix.lock changes.


MorganaFuture:
@v0idpwn Could you please take a look? Right now the _pull_request_target_ trigger means this audit always scans main, not the PR diff. This pr should  fix it

v0idpwn:
Merged different but very similar fix, thanks for this though!

Issue #239, opened 2023-12-25
Named prepared statements not working


marcustut:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Rust's [sqlx](https://github.com/launchbadge/sqlx) requires connection pooler to support named prepared statements in order to work properly, it has always been an issue up until the release of pgcat, that finally people can use it with a proper connection pooler, since #69 is closed I assume the latest release v1.0.0 for supavisor has been deployed for all Supabase users, however I did a test and sqlx still doesn't work and is due to named prepared statements.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

Clone the [supavisor-sqlx](https://github.com/marcustut/supavisor-sqlx) and replace the `DATABASE_URL` in `.env`, you should see this error. 

![image](https://github.com/supabase/supavisor/assets/59773847/8e4b4dd8-1bd9-40dd-9943-afb360398495)

## Expected behavior

Expect named prepared statements to work with sqlx

rbox-risk:
I think possibly related, but I too am using rust but with `tokio_postgres`. If I use existing direct connection to sbase everything works. As soon as I switch to connecting to the new connection pool via url to execute against a supabase function I get `Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E26000), message: "prepared statement \"s1\" does not exist", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("prepare.c"), line: Some(451), routine: Some("FetchPreparedStatement") }) }`

The error shows in both cases when I use `let stmt = client.prepare("...").await?;` then `client.query(&stmt,&params)` and when I bypass the `prepare` and straight to `client.query`. Left a comment as my code is part of a bigger thing and creating a repo might take a while - hence the comment. Thanks

Dkendal:
We're using Ecto with elixir and running into the exact same issue where the client is complaining the the number of parameters submitted differed from what was required.

salomartin:
This is a critical fix before you switch to Supavisor. Same issue when connecting from Kotlin based Spring apps.

marektomczyk:
Same for Postgraphile.

[...]

Issue #69, opened 2023-04-11
Support named prepared statements


chasers:
No pooler currently supports named prepared statements. Lots of SQL clients which use their own pool do e.g. Elixir's Ecto uses named prepared statements by default and ships with it's own pooling logic. 

It seems reasonable for a pooler to be able to support them. Named prepared statements can increase throughput significantly because the query plan is cached. 

Unnamed prepared statements do not use a cached query plan because it is deallocated with the next unnamed prepared statement. 

> If successfully created, a named prepared-statement object lasts till the end of the current session, unless explicitly destroyed. An unnamed prepared statement lasts only until the next Parse statement specifying the unnamed statement as destination is issued. [ref](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY)

## Questions

What if two different clients linked to one pg connection make two different prepared statements with the same name?

Because it's an async protocol can we add metadata to messages that Postgres will include coming back? Then we can use that to route the right messages to the right clients.

Can we use [pipelining](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-PIPELINING) to multiplex?

## Ideas

- Another pool mode
- Pin a client connection to a db conn such that each client would always get the same db conn from the pool

There would be some contention here probably because clients would have to wait on a specific conn but it would work. Question is does the contention negate the perf from the prepared statement. With another mode `transaction` pooling would work as expected but you could use this new mode if it works for your workload and provides perf gains.

hamiltop:
> What if two different clients linked to one pg connection make two different prepared statements with the same name?

To clarify this case: We generally _want_ both clients to use the same prepared statement. A classic "one process per request" infrastructure often means thousands of client connections and hundreds of server connections. Let's assume 10:1. If each of those processes handles 100 requests before getting cycled, then a given query used once in each request would use 5 custom plans and 95 generic plans for a 95% cache hit rate. If that prepared_statement is available to all clients using that server connection then we'd instead see a 99.5% hit rate.

chasers:
For sure. I think it really depends on how the client handles things though. Do they recreate the prepared statement on a connection if it doesn't exist? How do they name them? 

I'm thinking a separate pool mode which pins client connections to database connections would be good. So `transaction` mode behaves exactly like PgBouncer `transaction` mode and you could use this other mode if it's suitable for your workload and provides perf improvements.

hamiltop:
[...]

PR #980, opened 2026-05-12
perf(db_handler): use LRU eviction for backend prepared statements

MorganaFuture:

Performance improvement.

When the per-backend prepared statement set hits the 200-statement limit, eviction picks 40 entries at random (`Enum.take_random`) and sends `Close` for them. Hot statements get evicted with the same probability as cold ones,
so under churn we can drop a frequently used statement and pay for a fresh `Parse` round-trip on the next `Bind`.    Relates to #69 / #239.

## What is the new behavior?
   
Replaces the `MapSet`-backed set with a small LRU storage (`Supavisor.Protocol.PreparedStatements.BackendStorage`). Recency is bumped on `Parse` and on `Bind` for statements already known to the backend. Eviction now drops the least recently used statements first, so hot ones stay around.

No default behavior change still gated by the `named_prepared_statements` feature flag. Telemetry (`prepared_statements_evicted`) is unchanged. Covered by unit tests (basic API + edge cases) and a `StreamData` property test that checks the LRU ordering invariant across arbitrary `put`/`touch`/`delete` sequences.


v0idpwn (commented):
Thanks for the PR, @MorganaFuture, it looks great conceptually and the implementation is sound! I wanted to implement something like this but didn't get around to do it.

I'd like to suggest an architectural change, though: I'd like to make the db handler prepared statement storage backend pluggable. Why: we haven't really adopted this yet in production, and it would be nice to have both implementations at the same time so we can compare them in different scenarios. We could select it based on a feature flag, and eventually settle in a default. This also sets us up for a future in case we want to try a third storage.

WDYT?

MorganaFuture:
> Thanks for the PR, @MorganaFuture, it looks great conceptually and the implementation is sound! I wanted to implement something like this but didn't get around to do it.
> 
> I'd like to suggest an architectural change, though: I'd like to make the db handler prepared statement storage backend pluggable. Why: we haven't really adopted this yet in production, and it would be nice to have both implementations at the same time so we can compare them in different scenarios. We could select it based on a feature flag, and eventually settle in a default. This also sets us up for a future in case we want to try a third storage.
> 
> WDYT?

Great suggestion. BackendStorage is now a behaviour with two impls LRU and Random. Random is default.
Also could you please take a look at this pr [978](https://github.com/supabase/supavisor/pull/978). I am little annoyed by the failing CI/CD. Let's fix it.

lib/supavisor/protocol/prepared_statements/backend_storage.ex
```diff
@@ -0,0 +1,71 @@
...
+  @type t() :: struct()
+  @type name() :: PreparedStatements.statement_name()
+
+  @callback new() :: t()
+  @callback size(t()) :: non_neg_integer()
+  @callback member?(t(), name()) :: boolean()
+  @callback put(t(), name()) :: t()
+  @callback touch(t(), name()) :: t()
+  @callback delete(t(), name()) :: t()
+  @callback pop_oldest(t(), pos_integer()) :: {[name()], t()}
```
v0idpwn: Nitpick: for `Random`, these aren't the oldest, so it should probably be called `pop_n`

v0idpwn (approved):
Looks great, just one small note!

v0idpwn:
Regarding #978, we will do the bump separately as it requires appups to ship these version upgrades with zero downtime in production. These appups can be a bit nasty to write and test (and potentially not possible), so we will take our time to explore the version bumping in the next couple days.

lib/supavisor/protocol/prepared_statements/backend_storage.ex
v0idpwn: Or maybe `evict`

v0idpwn: gentle nudge @MorganaFuture, then I can merge!

MorganaFuture: @v0idpwn Done

MorganaFuture:
@v0idpwn  So let's merge?

v0idpwn:
Just wanted to thank you for this one @MorganaFuture, sorry for how long it took to merge! 💚

PR #983, opened 2026-05-13
chore: report current, permanent, base and previous versions

mentels:
Based on the `:release_handler.which_releases()`.

Versions:
- **current** — version currently running (`Application.spec(:supavisor, :vsn)`)
- **permanent** — version that boots on next restart (`:release_handler.which_releases/0` status `:permanent`)
- **base** — oldest `:old` release (the original installation, i.e. `upgrade_from`)
- **previous** — newest `:old` release (the version just before the last hot upgrade)

`base` and `previous` are `""` when no upgrade has been applied.

## Tag semantics

```
┌───────────────────────────────────────┬─────────┬───────────┬───────┬──────────┐
│               Scenario                │ current │ permanent │ base  │ previous │
├───────────────────────────────────────┼─────────┼───────────┼───────┼──────────┤
│ Steady state (no upgrades)            │ 2.9.4   │ 2.9.4     │ ""    │ ""       │
├───────────────────────────────────────┼─────────┼───────────┼───────┼──────────┤
│ After 1 upgrade (2.9.3→2.9.4)         │ 2.9.4   │ 2.9.4     │ 2.9.3 │ 2.9.3    │
├───────────────────────────────────────┼─────────┼───────────┼───────┼──────────┤
│ After 2 upgrades (2.9.2→2.9.3→2.9.4)  │ 2.9.4   │ 2.9.4     │ 2.9.2 │ 2.9.3    │
├───────────────────────────────────────┼─────────┼───────────┼───────┼──────────┤
│ During upgrade, before make_permanent │ 2.9.4   │ 2.9.3     │ 2.9.2 │ 2.9.3    │
└───────────────────────────────────────┴─────────┴───────────┴───────┴──────────┘
```

## Example output

Right after `install_release` (before `make_permanent`): `permanent` still points to the previous version, confirming the upgrade is in progress:

```
supavisor_prom_ex_application_version_info{...,base="",current="2.9.5",permanent="2.9.4",previous=""} 1
```

After `make_permanent`: both `current` and `permanent` are the new version, `previous` records what was just replaced:

```
supavisor_prom_ex_application_version_info{...,base="2.9.4",current="2.9.5",permanent="2.9.5",previous="2.9.4"} 1
```

Note: both entries appear simultaneously until the next upgrade, because Peep does not evict `last_value` entries whose tag combination is no longer emitted.

## Hot upgrade behaviour

The metric is emitted by a polling process inside `Supavisor.Monitoring.PromEx`. Because `Supavisor.Application.start/2` does not re-run on a hot upgrade, new polling processes added by an upgrade are not started automatically. `Supavisor.HotUpgrade` handles this by restarting `Supavisor.Monitoring.PromEx` as the final step of the upgrade — after all `appup` instructions have run so that `Application.spec(:supavisor, :vsn)` is already current on the first poll.

### Risks of the PromEx restart

**One `:persistent_term` GC sweep — latency spike.** Peep stores its metric index (event→metric mappings, ETS storage references) in `:persistent_term`. Replacing or erasing a key there triggers a global GC pass across every process on the node, because all processes hold read references that must be invalidated. The restart causes exactly one such sweep, though when it fires depends on the Peep version:

- **Current version (buggy `erase`):** `Peep.Persistent.erase` calls `:persistent_term.erase(name)` using the bare atom instead of the actual key `{Peep, name}` — a no-op that triggers no GC. The stale `{Peep, name}` entry survives termination and is only replaced when the new Peep starts and calls `:persistent_term.put({Peep, name}, new_term)`, replacing the existing key — **GC fires on `restart_child`**. Bug fixed in Peep v5.0.1: [47080c5](https://github.com/rkallos/peep/commit/47080c553ae1b3b39efdf3bb743d6a4fb4691a3d).
- **Fixed version:** `erase` correctly removes `{Peep, name}` — **GC fires on `terminate_child`**. The subsequent `put` inserts a fresh key and triggers no GC.

**Brief handler detachment gap — metric data loss.** After `terminate_child`, telemetry handlers for all Peep-managed events are still registered (detachment via `EventHandler.detach` happens inside `Peep.terminate/2` but only after `erase`). Any `:telemetry.execute` fired during this window runs the handler in the caller's process. In the current buggy version the stale `{Peep, name}` entry is still present, so `Peep.Persistent.fetch` returns the stale struct — the pattern match succeeds, but the subsequent ETS write to the now-destroyed storage tables raises `ArgumentError`. In the fixed version `fetch` returns `nil` and the failure is a pattern match error. Either way the exception is caught by telemetry, the handler is detached, and collection resumes once `restart_child` completes. The data lost is limited to events fired during the restart window.

**Calling processes are not affected.** The handler function runs in the process that called `:telemetry.execute`, not in Peep. The exception it raises is caught inside telemetry's own `do_execute` loop (the `try/catch` wrapping every handler call in `telemetry.erl`), which detaches the failing handler and fires a `[telemetry, handler, failure]` diagnostic event, then continues. The exception never propagates back to the caller.

---

Once this is merged, we need to do a Grafana PR.


lib/supavisor/monitoring/cluster.ex
```diff
@@ -48,6 +63,33 @@ defmodule Supavisor.PromEx.Plugins.Cluster do
     :telemetry.execute([:supavisor, :prom_ex, :cluster], %{size: cluster_size})
   end
 
+  @spec emit_app_version() :: :ok
+  def emit_app_version do
+    current = Application.spec(:supavisor, :vsn) |> to_string()
+    releases = :release_handler.which_releases()
```
v0idpwn, line 69: TIL :release_handler.which_releases()

v0idpwn (approved):

v0idpwn (approved):

filipecabaco (approved):

PR #993, opened 2026-05-21
fix: chunk TLS sends to work around libpq async deadlock

v0idpwn:
libpq's async API (PQconsumeInput/PQisBusy) can hang when a TLS record
leaves bytes in OpenSSL's userspace buffer that pqReadData doesn't
drain in a single SSL_read: the client then polls the socket for data
that has already arrived. Capping our TLS records at libpq's 8KB
initial input buffer avoids producing the framing that triggers it.

See https://www.postgresql.org/message-id/flat/57d1e8b1-d016-4cea-9b60-63cbdb40eb81%40iki.fi


lib/supavisor/db_handler.ex
```diff
@@ -894,18 +892,53 @@ defmodule Supavisor.DbHandler do
...
+  defp client_send(%{client_sock: sock}, payload) do
+    HandlerHelpers.sock_send(sock, payload)
+  end
+
+  defp chunk_send([], _sock), do: :ok
+
+  defp chunk_send(iovec, sock) do
+    {chunk, rest} = take_chunk(iovec, @tls_send_chunk_size, [])
+
+    case HandlerHelpers.sock_send(sock, chunk) do
```
v0idpwn, line 925: What's annoying is that this is inefficient and will cause us to do way more message passing and sys calls than before. Thinking if we should gate this behind a connection option of some sort.

mentels: That could be a good idea; that could be a document option and only those who run into the issue would turn this behaviour on.

mentels (approved):
LGTM, though I would not ship this without proper tests.

PR #1001, opened 2026-05-22
chore: 2.9.6 release with soft_purge fix for hot-path modules and lowered suspend timeout for DbHandler

mentels:
Bumps to 2.9.6-rc.0 and switches `ClientHandler`, `DbHandler`, `ClientHandler.Error`, `ClientHandler.Checks` to `soft_purge` (PostPurge) in the relup appups.

Under load, the deferred `code:purge/1` inside `release_handler:make_permanent/1` kills ClientHandlers blocked in `:gen_statem.call(db_handler, …)` during the suspended-DbHandler window — they have `Supavisor.ClientHandler.*` frames on their stack and count as "lingerers". `soft_purge` makes `make_permanent` skip the kill; both code versions stay resident until the next relup drains the old one. Suspend/resume kept; all other modules stay `brutal_purge`.

The 2.9.5 relups are also patched in-place so a to-2.9.5 upgrade with the fix remains possible.

Also fixes `mix.exs:151` so the appup-copy step survives hyphenated to-versions: the filename split now uses `parts: 2`, otherwise `supavisor-2.9.6-rc.0.appup` would parse as `["supavisor", "2.9.6", "rc.0"]` and crash the release build with a `MatchError`.

Plan: tag this branch as v2.9.6-rc.0, validate in staging, then a follow-up commit bumps to v2.9.6.


v0idpwn (commented):
~~Won't this cause make_permanent to fail if we have processes running old code?~~

Edit: sorry, realized this doesn't make sense

relups/2.9.0-2.9.5/appups/supavisor-2.9.5.appup
```diff
@@ -10,9 +10,10 @@
     {delete_module,'Elixir.Supavisor.Logger.Filters'},
     {load_module,'Elixir.Jason.Encoder.SupavisorWeb.OpenApiSchemas.Tenant',[]},
     {load_module,'Elixir.Jason.Encoder.SupavisorWeb.OpenApiSchemas.User',[]},
-    {suspend,['Elixir.Supavisor.DbHandler']},
+    {suspend,[{'Elixir.Supavisor.DbHandler',50}]},
     {load_module,'Elixir.Supavisor.CircuitBreaker',[]},
-    {load_module,'Elixir.Supavisor.ClientHandler.Checks',[]},
+    {load_module,'Elixir.Supavisor.ClientHandler.Checks',
+                 soft_purge,soft_purge,[]},
```
v0idpwn: Should probably be `brutal_purge,soft_purge` for all of them

v0idpwn (approved):

PR #1007, opened 2026-05-22
chore(ci): allow manual dispatch of stage deploy workflow

mentels:
Adds `workflow_dispatch` with inputs (`upgrade_from`, `upgrade_previous`, `upgrade_to`, `regions`, `type`) so a stage deploy can be triggered from the Actions UI without editing `staging.config` and pushing.

The `Set upgrade variables` step now branches on `github.event_name`: inputs when dispatched manually, `staging.config` on push (existing behaviour preserved). Also aborts up-front when `type=soft` and `upgrade_previous == upgrade_to`, which otherwise silently hangs the second `mix release` on \`Release already exists. Overwrite? [Yn]\` and surfaces only as a confusing \`mv: cannot stat ...tar.gz\` later (see failed run [#26285657935](https://github.com/supabase/supavisor/actions/runs/26285657935)).

No change to soft/hard branching, action versions, or any other step.

## Test plan
- [ ] Merge, then trigger from Actions UI ("Run workflow") with `upgrade_from=2.9.4`, `upgrade_previous=…`, `upgrade_to=2.9.6-rc.0`, `regions=all`, `type=soft`.
- [ ] Confirm next push to `release_stage` (existing flow) still reads `staging.config` and works as before.

### DIFF
--- a/.github/workflows/stage.yml
+++ b/.github/workflows/stage.yml
@@ -52,16 +52,25 @@ jobs:
           otp-version: "27.2.1"
           elixir-version: "1.18.2"
       - name: Set upgrade variables
+        env:
+          # Pass dispatch inputs via env (not direct ${{ }} interpolation) to
+          # avoid shell injection. See
+          # https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable
+          INPUT_FROM: ${{ inputs.upgrade_from }}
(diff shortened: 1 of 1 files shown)


deepthi (approved):

v0idpwn (approved):

mentels:
Tested successfully for both [hard](https://github.com/supabase/supavisor/actions/runs/26502862794) and [soft](https://github.com/supabase/supavisor/actions/runs/26579270020) deployments.

PR #1026, opened 2026-06-06
test: fix race in tenant metrics aggregation tests

Snehil-Shah:

Adds a simple and defensive synchronization step to wait for all tenant registrations to happen before executing metrics.

https://github.com/supabase/supavisor/actions/runs/26826883951/job/79096879479

We don't wait for all `Registry.register` calls to finish before executing metrics aggregation. Generally fine as they usually finish on time. But still worth a defensive check as the above run failed.

## What is the new behavior?

Waits for all `Registry.register` calls to finish by sending a registered event to the parent test process that we count for.


v0idpwn (approved):

v0idpwn:
Thank you! 💚

PR #1027, opened 2026-06-06
fix: auto-detect ipv6 from db hostname

### LINKED ISSUE #314: Failed to connect using interval private network Postgres
I've set up a private network with PostgreSQL on both Fly.io and Railway.app, but I'm encountering issues when trying to connect Supabase to PostgreSQL via the internal network. The error message I receive is: `[error] Postgrex.Protocol (#PID<0.2843.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (postgres.*.internal:5432): non-existing domain - :nxdomain`. Interestingly, when I switch to using the regular domain for the public network, the connection works flawlessly. I have verified that the internal network functions correctly with other services. To troubleshoot, I attempted to modify the database configuration by adding a socket option `socket_options: [:inet6]`, yet the issue persists with the domain not being recognized. This is the only service I'm having trouble with when connecting through the private network. Given my limited knowledge of Elixir, I find it quite challenging to debug this issue further. Does anyone have any insights or suggestions?

Snehil-Shah:

Updates runtime config to auto-detect IP version from the metadata DB hostname, so it can work out of the box for ipv6 hosts, achieving parity with how we handle tenant DBs, and for someone migrating from pgbouncer without having them touch new knobs.

Supavisor couldn't start up when its `DATABASE_URL` pointed at an IPv6-only hostname (like fly.io's internal dns). It would fail on `:nxdomain` errors.

The IP version is configurable via the `SUPAVISOR_DB_IP_VERSION` env var (defaults to ipv4), but ideally can be auto-detected (what this PR aims to do).

## What is the new behavior?

We already had the necessary helper that we used to resolve tenant DBs. Reused the same to set for our metadata db as well. I have tested and verified the behavior on some test deployments on fly.io with a postgres instance on a `.internal` hostname. It works.

## Additional context

Closes #314


v0idpwn (approved):

v0idpwn:
Thanks! 💚

PR #1029, opened 2026-06-08
fix: add backend_pid to logger metadata allow-list

v0idpwn:
The DbHandler process added the postgres backend PID to metadata, but this wasn't available in the logs due to lacking configuration.

POOLER-180


v0idpwn:
Verified not necessary.

PR #1032, opened 2026-06-09
fix(auth): use secure_compare for credential comparisons

Snehil-Shah:

Nitpicky change to use `secure_compare` when comparing client-fed credentials. We are already using this in https://github.com/supabase/supavisor/blob/main/lib/supavisor/client_handler/auth_methods/password.ex#L86, so it closes the inconsistency as well.


lib/supavisor/client_handler/auth_methods/password.ex
```diff
@@ -73,7 +73,7 @@ defmodule Supavisor.ClientHandler.AuthMethods.Password do
   end
 
   defp validate_password(password, %{tenant: %{require_user: true}} = context) do
-    if password == context.user.db_password do
+    if Plug.Crypto.secure_compare(password, context.user.db_password) do
```
v0idpwn, line 76: 👍🏻 here

lib/supavisor/client_handler/auth_methods/password.ex
```diff
@@ -96,7 +96,7 @@ defmodule Supavisor.ClientHandler.AuthMethods.Password do
         client_key = :pgo_scram.hmac(salted_password, "Client Key")
         computed_stored_key = :pgo_scram.h(client_key)
 
-        if computed_stored_key == sasl_secrets.stored_key do
+        if Plug.Crypto.secure_compare(computed_stored_key, sasl_secrets.stored_key) do
```
v0idpwn: I don't think this has any risk of timing attack? The value is hashed.

lib/supavisor/client_handler/auth_methods/scram.ex
```diff
@@ -134,7 +134,7 @@ defmodule Supavisor.ClientHandler.AuthMethods.SCRAM do
   defp validate_scram_proof(context, client_proof) do
     client_key = :crypto.exor(Base.decode64!(client_proof), context.signatures.client)
 
-    if Helpers.hash(client_key) == context.secret.stored_key do
+    if Plug.Crypto.secure_compare(Helpers.hash(client_key), context.secret.stored_key) do
```
v0idpwn: Same as above here

lib/supavisor/helpers.ex
```diff
@@ -295,7 +295,7 @@ defmodule Supavisor.Helpers do
 
     client_key = :pgo_scram.hmac(salted_password, "Client Key")
     stored_key = :pgo_scram.h(client_key)
-    stored_key == secrets.stored_key
+    Plug.Crypto.secure_compare(stored_key, secrets.stored_key)
```
v0idpwn: And here

lib/supavisor/client_handler/auth_methods/password.ex
Snehil-Shah: @v0idpwn Yeah, you are right. I added secure_compare to all the places where you have some client-controlled input. Does it hurt much to be extra defensive? WDYT?

Snehil-Shah: I'll remove the three sites tbh. Plain text comparisons is the only real risk 🤔

Snehil-Shah: Thinking out loud, they can reproduce the hash value locally too. So they essentially know what bytes are being compared at this line. So yeah, guessing the first byte of the stored key is possible. Although, guessing the next bytes would start to get impossible as many inputs can produce the same first few bytes of a hash.🤔🤔

v0idpwn: fwiw, I just verified Postgres source as reference, and its SCRAM implementation uses `memcmp` for these hash comparisons, so I think we are good to go as it is

v0idpwn (approved):

v0idpwn:
Thank you! 💚

PR #1036, opened 2026-06-14
fix(db_handler): handle ReadyForQuery split across TCP reads

Snehil-Shah:

Uses the `MessageStreamer` via the backend message handler to detect ReadyForQuery, so it's always detecting from a framed TCP stream, and not a single packet in which the message is prone to be split.

ReadyForQuery appears at the end of a variable-sized result message in the hot path. So there is a risk of the ReadyForQuery message splitting across two socket reads due to a network hiccup, or just because the OS decided to. This wouldn't clean up the DB handler even if it's done with use.

## What is the new behavior?

Uses the existing Backend Message handler to use the stream state and detect ReadyForQuery even across TCP packets.

### DIFF
--- a/lib/supavisor/db_handler.ex
+++ b/lib/supavisor/db_handler.ex
@@ -407,6 +407,41 @@ defmodule Supavisor.DbHandler do
   end
 
   # forward the message to the client
+  def handle_event(
+        :info,
+        {proto, _, bin},
+        :busy,
+        %{backend_message_streaming: true, caller: caller} = data
+      )
+      when is_pid(caller) and proto in @proto do
+    Logger.debug("DbHandler: Got messages: #{Debug.packet_to_string(bin, :backend)}")
+
+    data = reset_ready_for_query(data)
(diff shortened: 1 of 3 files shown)


Snehil-Shah:
@v0idpwn I noticed a few inconsistencies around how packets are read. A few of the handlers (in both `client_handler` and `db_handler`) don't care for this for small packets, which is fine, as there isn't much risk of them splitting. But there is also a handler for `:waiting_cleanup` that has its own separate machinery for accumulating packets into its own buffer instead of reusing `MessageStreamer`. I was just curious if there are any reasons around this, or if this is just historical inconsistencies piling up?

v0idpwn:
Hi, @Snehil-Shah. Yeah, we have been (slowly) migrating towards the safer packet handling approach, and I try to follow it in new code.

The buffering is done differently in the cleanup as the message handler was still behind a feature flag. 

I'm now more confident in the message handler and intend to switch more places to use it.

Snehil-Shah:
Closing in favour of #1079.

PR #1037, opened 2026-06-16
feat: forward client application_name in session mode

### LINKED ISSUE #343: Prepend application_name from client to application_name set by DbHandler only in session mode

Snehil-Shah:

A simple cosmetic, quality-of-life, better debuggability change. Names the specific backend connection to the client's application name.

Name is always "Supavisor". Makes sense in transaction mode.

## What is the new behavior?

Name is now what the client provides (in session mode). Makes sense in session-modes where we want to give them the feeling that they own the backend connection.

## Additional context

Closes #343

### DIFF
--- a/lib/supavisor/db_handler.ex
+++ b/lib/supavisor/db_handler.ex
@@ -452,8 +452,11 @@ defmodule Supavisor.DbHandler do
       # Only when checked out to this client (:busy).
       state == :busy ->
         Logger.debug("DbHandler: Setting application_name to #{inspect(name)}")
-        query = "SET application_name = #{quote_literal(name)}"
-        :ok = HandlerHelpers.sock_send(data.sock, :pgo_protocol.encode_query_message(query))
+
+        query =
+          Server.extended_query("SELECT set_config('application_name', $1, false)", [name])
+
+        :ok = HandlerHelpers.sock_send(data.sock, query)
 
         {:next_state, :setting_application_name,
          Map.merge(data, %{set_app_name_from: from, pending_bin: <<>>}),
@@ -930,9 +933,6 @@ defmodule Supavisor.DbHandler do
       data
     end
   end
-
-  # Quote a value as a SQL string literal.
-  defp quote_literal(value), do: "'" <> String.replace(value, "'", "''") <> "'"
 
   # If the prepared statement exists for us, it exists for the server, so we just send the
   # bind to the socket. If it doesn't, we must send the parse pkt first.
--- a/lib/supavisor/db_handler.ex
+++ b/lib/supavisor/db_handler.ex
@@ -58,6 +58,7 @@ defmodule Supavisor.DbHandler do
           | :busy
           | :terminating_with_error
           | :waiting_for_secrets
+          | :setting_application_name
 
   @sock_closed [:tcp_closed, :ssl_closed]
   @proto [:tcp, :ssl]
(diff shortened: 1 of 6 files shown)


lib/supavisor/db_handler.ex
```diff
@@ -451,6 +463,61 @@ defmodule Supavisor.DbHandler do
...
+      true ->
+        {:keep_state_and_data, {:reply, from, :ok}}
+    end
+  end
+
+  # Swallow the SET application_name response so it never reaches the client.
+  # Setting application name should feel like an implicit feature.
+  def handle_event(:info, {proto, _, bin}, :setting_application_name, data)
+      when proto in @proto do
+    buffered_bin = data.pending_bin <> bin
```
Snehil-Shah, line 473: I pretty much based the entire thing on the `:waiting_cleanup` mechanism, including the `pending_bin` accumulation logic. It would be cleaner if #1036 is merged first, so we can reuse the ReadyForQuery detection using the `MessageStreamer` for consistency moving forward.

v0idpwn (commented):
I like the concept! 

But it conflicts with #891. Could you please target it instead? I think the two approaches could be merged so that in session mode you always get the application name and in txn mode you never get. WDYT?

v0idpwn:
>Name is always "Supavisor". Makes sense in transaction mode.

It doesn't... It's currently a mess, hence #891.

Snehil-Shah:
@v0idpwn Yes, I have rebased and now it targets your branch.

> > Name is always "Supavisor". Makes sense in transaction mode.
>
> It doesn't... It's currently a mess, hence 891.

I meant, the application_name on PG connection is always hardcoded to `Supavisor` (except in proxy mode), right?

v0idpwn:
>I meant, the application_name on PG connection is always hardcoded to Supavisor (except in proxy mode), right?

It attempts to use app_name if it was provided, but it uses the one for the first connection in the pool, which kinda sucks :)

lib/supavisor/db_handler.ex
```diff
@@ -427,6 +439,61 @@ defmodule Supavisor.DbHandler do
...
+         Map.merge(data, %{set_app_name_from: from, pending_bin: <<>>}),
+         {:state_timeout, 5_000, :set_application_name_timeout}}
+
+      true ->
+        {:keep_state_and_data, {:reply, from, :ok}}
+    end
+  end
+
+  # Swallow the SET application_name response so it never reaches the client.
+  # Setting application name should feel like an implicit feature.
```
v0idpwn:
```suggestion
```

lib/supavisor/db_handler.ex
```diff
@@ -427,6 +439,60 @@ defmodule Supavisor.DbHandler do
...
+        )
+
+        {:keep_state_and_data,
+         {:reply, from, {:error, :set_application_name_not_supported_in_transaction_mode}}}
+
+      # Only when checked out to this client (:busy).
+      state == :busy ->
+        Logger.debug("DbHandler: Setting application_name to #{inspect(name)}")
+        query = "SET application_name = #{quote_literal(name)}"
+        :ok = HandlerHelpers.sock_send(data.sock, :pgo_protocol.encode_query_message(query))
```
v0idpwn: I'm a bit afraid of SQLi here (or even just breaking the protocol, or emitting invalid SQL). Can we perhaps use the extended query protocol? Or we will need to improve the escaping.

Snehil-Shah: @v0idpwn Yes, we can use `SELECT set_config('application_name', $1, false)`. One thing, though, for writing the extended query, `:pgo_protocol.encode_bind_message/5` requires a pgo connection. Can I go ahead and make a new simple shared helper in the `Supavisor.Protocol.Server` module for building the binary for Bind messages with text params? (or is there any existing machinery I'm missing out on?)

v0idpwn: I think we can have a function that generates a full query: Parse/Prepare/Execute/Sync on a binary. We send this and we discard the output until the Z message is received.

Snehil-Shah: @v0idpwn Done in https://github.com/supabase/supavisor/pull/1037/changes/e6be5c8c43bbe5c6f727c694ace5fc288bdb201d. The call site feels pretty clean now. WDYT?

v0idpwn: Looks great to me!

PR #1038, opened 2026-06-18
fix: always use ssl for upstream with jit

staaldraad:

Bug fix

If `ssl_enforce` is not configured and a client tries to connect with `jit`, supavisor crashes.

## What is the new behavior?

Supavisor will follow the client's TLS connection upstream, but use `verify_none` if `ssl_enforce` is not set


ChasNelson1990:
Blocker for me currently, anything I can do to help it get merged?

v0idpwn:
Hey, @ChasNelson1990, if you are having issues with JIT in your project, please enable requiring TLS alongside the JIT. We are adding a gate to prevent JIT from being enabled without TLS being required.

PR #1060, opened 2026-07-02
Handle client socket errors on DbHandler, sweep zombie ports, hold check-in till flush complete

v0idpwn:
When the timeout closes ({send_timeout_close, true} (a ranch default)) are triggered in the inet backend, the underlying TCP socket is closed, but there's no notification to the owner process, and the port stays alive.

We must then make sure that when we get the timeout error from send, we adequately react and the relevant processes are terminated. The DbHandler didn't check the return value when writing to the client, and when timing out would cause zombie ports+client handlers, which could even hold a database session, and potentially create long transactions. I reproduced this scenario in a test.

This doesn't apply to the upstream sockets to database, which don't use this option.

The dead port sweeper is cherry picked from https://github.com/supabase/supavisor/pull/1056, which was the bug mitigation targeted at the current v2.9 release.

Also addresses a related race condition that could cause a DbHandler that is busy flushing data to a slow client to be checked out by another client if (and only if) the query had already completed from the DbHandler perspective. This was fixed by making the DbHandler checkin itself when its work is complete, instead of relying on the ClientHandler to check it in.I don't expect this to make much of a difference in real life, but its a more correct behaviour.

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -515,7 +515,9 @@ defmodule Supavisor.ClientHandler do
   def handle_event(:cast, {:db_status, :ready_for_query}, :busy, data) do
     Logger.debug("ClientHandler: Client is ready")
 
-    db_connection = maybe_checkin(data.mode, data.pool, data.db_connection)
+    # In transaction mode the DbHandler checks itself back into the pool once
+    # it finishes forwarding the response; we only drop our reference to it.
+    db_connection = if data.mode == :transaction, do: nil, else: data.db_connection
 
     {_, stats} =
       if data.local,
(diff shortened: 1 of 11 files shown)


lib/supavisor/client_handler.ex
```diff
@@ -515,7 +515,9 @@ defmodule Supavisor.ClientHandler do
   def handle_event(:cast, {:db_status, :ready_for_query}, :busy, data) do
     Logger.debug("ClientHandler: Client is ready")
 
-    db_connection = maybe_checkin(data.mode, data.pool, data.db_connection)
+    # In transaction mode the DbHandler checks itself back into the pool once
+    # it finishes forwarding the response; we only drop our reference to it.
+    db_connection = if data.mode == :transaction, do: nil, else: data.db_connection
```
mentels, line 530: If we don't unlink here from the DbHandler, if it crashed, it might potentially crash this connection - eg. when it sits idle waiting for another query.

mentels: Now I see we unlink in DbHandler....but it doesn't change the situation; still a crashed DbHandler tears down this connection which might be using a different worker already which feels weird; but I don't see a better approach here.

mentels (commented):
LGTM, there's one remark.

lib/supavisor/client_handler.ex
v0idpwn: If I understand correctly, this behaviour is desired, look at the following case:

1. Client sends a query
2. Db returns rows + ready_for_query
3. DbHandler reads the ready_for_query and sends the db_ready to the client handler
4. ClientHandler receives the db_ready message but doesn't unlink (new change in this PR)
5. DbHandler starts sending the data to the client socket
6. DbHandler gets a timeout sending the data to the client socket
7. DbHandler terminates
8. ClientHandler terminates due to the link

If we unlink at (4), the hanging client handler will remain up.

v0idpwn: I've explained a bit more of this issue on #1061

mentels (approved):

PR #1064, opened 2026-07-07
chore: rollback ap-southeast-2-sec to v2.9.0

mentels:
Standby rollback for supabase/supavisor#1062.

**Do NOT merge unless rolling back the ap-southeast-2-sec upgrade.**


mentels:
Not needed.

PR #1067, opened 2026-07-08
chore: rollback cac11 to v2.9.2

mentels:
Standby rollback for https://github.com/supabase/supavisor/pull/1065.

**Do NOT merge unless rolling back the cac11 upgrade.**


mentels:
Not needed

PR #1068, opened 2026-07-08
fix(client_handler): wire in client_idle_timeout

Snehil-Shah:

Bug fix: Fixes and wires in the broken `client_idle_timeout` functionality.

`client_idle_timeout` silently does nothing. Idle connections are never being closed. The client_handler gen_statem has been firing a mismatched timeout event since #377.

## What is the new behavior?

Fixed the mismatch and also decoupled the timer to a state timeout, which makes more sense semantically and avoids the messy mix-up with the heartbeat timer. The heartbeat timer keeps using the event timeout.

## Additional context

Also stitched a simple integration test for this, taking help from helpers in `protocol_integration_test.exs`.

### DIFF
--- a/test/integration/client_idle_timeout_test.exs
+++ b/test/integration/client_idle_timeout_test.exs
@@ -1,0 +1,106 @@
+defmodule Supavisor.Integration.ClientIdleTimeoutTest do
+  use Supavisor.DataCase, async: false
+
+  @moduletag :integration
+
+  @idle_timeout_ms 1000
+
+  setup do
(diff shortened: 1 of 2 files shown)


test/integration/client_idle_timeout_test.exs
Snehil-Shah, line 1: @v0idpwn I can see #1060 introduces some test helpers that we can use here too. Will update this file once that's in.

v0idpwn (approved):
Looks great!

Issue #300, opened 2024-02-07
Queries on tables that use CITEXT are 4x slower when using Supavisor vs a direct connection to the db


jensen:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

Originally searched on discord and posted there, the amazing garyaustin suggested I post the issue on the Supavisor GitHub. https://discord.com/channels/839993398554656828/1204584952029782026

I tested this issue with a local install of Supavisor following the instructions found here: [https://supabase.github.io/supavisor/](https://supabase.github.io/supavisor/development/installation/)

## Describe the bug

Queries to tables that use the CITEXT type are taking longer than queries on tables that do not use this type. Even if the query is `select * from table`. This bug can be described as weird because it does not seem to occur when connecting to the database directly.

| Connection | CITEXT | Duration |
|---|---|---|
| Supavisor | Yes | 1111ms |
| Supavisor | No | 277ms |
| Direct | Yes | 204ms |
| Direct | No | 139ms |

Background:

We setup a Supabase instance in eu-west-2, but are doing the development in us-west. This means that our queries have about **110ms** of extra latency which is acceptable. Recently we noticed that queries to the `users` table are taking about **1200-1500ms** while queries to other tables are taking less than **300ms**. We recently switched to the Supavisor pooler, and do not recall if the queries were slow when we were using PGbouncer.

## To Reproduce

1. Setup a new Supabase instance, the farther away the more pronounced the difference in duration.
2. Install CITEXT extension
3. Add a table that has a few columns that use the CITEXT type. In our case we have a users table with email, firstName and lastName that are of type CITEXT.
4. Add a table that has a few columns, but do not use the CITEXT type.
5. Connect with Prisma with `postgres://postgres.tenant:password@region.supabase.com:6543/postgres?pgbouncer=true
6. Query from the first table with something like `select * from users` and record the duration.
7. Query from the second table with something like `select * from centres` and record the duration.

## Expected behavior

The queries should take the same duration when using the connection pooler.

## Logs

Connection to Supavisor eu-west-2 (no CITEXT):

```
[...]

PR #1074, opened 2026-07-09
docs: explain citext + pgbouncer=true slowdown and its remedy

abidedavana:
I looked into #300 and the short version is this isn't a Supavisor bug. The
slowdown is on Prisma's side, and it happens even with no pooler in the path.

Refs #300

Documentation only. No change to proxy behavior.

If you've got `citext` columns (or any extension type) and you're using Prisma
with `pgbouncer=true`, queries in transaction mode are way slower than the same
table with plain `text` columns that's what #300 is about. The docs don't
explain it, and the FAQ still says prepared statements only work in session
mode, which isn't true anymore.

Here's what's actually going on. `citext`'s type OID gets assigned when you run
`CREATE EXTENSION`, so unlike `text` the client doesn't know it up front. With
`pgbouncer=true`, Prisma throws away its statement and type caches on every
transaction (`DEALLOCATE ALL`), so it has to look the OID back up with
`SELECT ... FROM pg_type`  once per citext column, every single transaction.
Add latency and those round trips are basically the whole query time. I
confirmed it's not us two ways: it reproduces hitting Postgres directly with no
pooler, and profiling the client/db handlers with `:eprof` during the slow
query shows under 1% of the time is actually spent in Supavisor. We're just
passing the round trips through.

## What is the new behavior?

The fix is already in Supavisor, it just wasn't documented  turn on the
`named_prepared_statements` flag and drop `pgbouncer=true`, and Prisma keeps its
caches so the lookups happen once instead of every transaction. Session mode
works too.

- `docs/orms/prisma.md` documents the flag as the alternative to
  `pgbouncer=true`, plus a section on why extension-typed columns get slow and
  how to avoid it.
- `docs/faq.md`  fixes the stale "session mode only" answer.
- `test/integration/prepared_statements_test.exs` — new test that prepares and
  runs a statement returning a citext column through transaction mode.
- `priv/repo/seeds_after_migration.exs` — sets up the citext extension and a
  test table for it. (The `supabase_admin` role bit is there because the image's
  extension scripts expect that role and the CI compose setup skips the init
  that'd create it.)

Numbers from a local repro, ~70ms round-trip latency, Prisma 6.19.3, 3 citext
columns (p50):

| setup | citext | text |
|---|---|---|
| transaction + `pgbouncer=true` (before) | 796ms | 363ms |
| transaction + `named_prepared_statements`, no `pgbouncer=true` (after) | 82ms | 79ms |
| direct connection | 74ms | — |

## Additional context

Those numbers are from a local toxiproxy setup with injected latency, not a real
WAN, but they line up with what's reported in #300 and the round-trip counts
don't depend on latency anyway. Full `mix test` passes the same as the base
branch (the handful of failures are pre-existing cluster/metrics stuff,
unrelated); the new citext test passes.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -12,7 +12,11 @@ subsequently should show up in your exception monitoring software.
 
 ## Does Supavisor support prepared statements?
 
-It currently supports prepared statements only in session mode.
+It supports prepared statements in session mode. In transaction mode, named
+prepared statements are supported when the `named_prepared_statements`
+feature flag is enabled, either globally with the
+`NAMED_PREPARED_STATEMENTS_ENABLED` environment variable or per tenant via
+`feature_flags`.
 
 ## Why do you route connections to a single Supavisor node when deployed as a cluster?
 
--- a/docs/orms/prisma.md
+++ b/docs/orms/prisma.md
@@ -15,6 +15,27 @@ string with Prisma.
 
 The `pgbouncer=true` connection string parameter is compatible with Supavisor.
 
+Alternatively, Supavisor supports named prepared statements in transaction
+mode when the `named_prepared_statements` feature flag is enabled, either
+globally with the `NAMED_PREPARED_STATEMENTS_ENABLED` environment variable or
+per tenant via `feature_flags`. With it enabled, Prisma can connect through
+transaction mode without `pgbouncer=true`.
+
+## Slow Queries on Tables with Extension Types
+
+With `pgbouncer=true`, Prisma does not reuse its per-connection statement and
+type caches across transactions. For columns whose types come from extensions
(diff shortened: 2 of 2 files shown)


v0idpwn (commented):
Thanks for the investigation!

Happy to merge the documentation change, but I don't think the test is testing anything in the pooler, can we drop it?

abidedavana:
Makes sense, that test wasn't really exercising the pooler, since Supavisor doesn't do anything type-specific. yea sure we can drop it

v0idpwn (approved):

v0idpwn:
Thanks again, @abidedavana! 💚

abidedavana:
> Thanks again, @abidedavana! 💚

Happy to help,  : )

PR #1077, opened 2026-07-10
feat(monitoring): add metric for AMI version

terrywilcox:
Expose the node's AMI version as a Prometheus info metric (supavisor_prom_ex_ami_version_info) with the version carried in a `version` label, mirroring the existing app-version metric pattern. The value is read from the AMI_VERSION env var, defaulting to "" when unset.

Feature

Doesn't put AMI version in metrics

## What is the new behavior?

Add AMI version to metrics

## Additional context

https://linear.app/supabase/issue/POOLER-554/add-metric-for-ami-version

### DIFF
--- a/lib/supavisor/monitoring/cluster.ex
+++ b/lib/supavisor/monitoring/cluster.ex
@@ -52,10 +52,15 @@ defmodule Supavisor.PromEx.Plugins.Cluster do
             tags: [:current, :permanent, :base, :previous]
           )
         ]
-      ),
-      Polling.build(
-        :supavisor_ami_version_events,
-        poll_rate,
+      )
+    ]
+  end
+
+  @impl true
+  def manual_metrics(_opts) do
+    [
+      Manual.build(
+        :supavisor_ami_version_manual_metrics,
         {__MODULE__, :emit_ami_version, []},
         [
           last_value(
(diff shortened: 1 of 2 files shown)


lib/supavisor/monitoring/cluster.ex
```diff
@@ -52,6 +52,20 @@ defmodule Supavisor.PromEx.Plugins.Cluster do
             tags: [:current, :permanent, :base, :previous]
           )
         ]
+      ),
+      Polling.build(
```
v0idpwn: Should we use a longer poll interval for this one since it doesn't change? 🤔

lib/supavisor/monitoring/cluster.ex
```diff
@@ -90,6 +104,17 @@ defmodule Supavisor.PromEx.Plugins.Cluster do
     )
   end
 
+  @spec emit_ami_version() :: :ok
+  def emit_ami_version do
+    version = System.get_env("AMI_VERSION") || ""
```
v0idpwn: I wonder if we should not emit it when its not there. For example, self-hosted deployments usually won't set up this env var.

terrywilcox: That seems reasonable. I'll change it.

lib/supavisor/monitoring/cluster.ex
terrywilcox: It does seem overkill to publish the same number every 15 seconds. 
An AMI version change means a restart, which triggers a poll when it starts up? We can't miss any value changes?

v0idpwn: Yep! I know there's something in PromEx for "publish once" metrics, maybe we can use that?

v0idpwn: It's used for application versions/elixir version/otp version

terrywilcox: Changed it to manual rather than poll.

lib/supavisor/monitoring/cluster.ex
terrywilcox: Does not emit if not set.

v0idpwn (approved):

PR #1079, opened 2026-07-11
fix: wait for all DB replies before releasing backend

### LINKED ISSUE #1061: [RFC]: Pipelined Transactions
When a client [pipelines](https://www.postgresql.org/docs/current/libpq-pipeline-mode.html) multiple transactions in a single batch, replies beyond the first few transactions are dropped, and the backend is released early, causing the client to just hang waiting till timeout. (in transaction mode)

Client libraries like [postgres.js](https://github.com/porsager/postgres) and [tokio-postgres](https://docs.rs/tokio-postgres/latest/tokio_postgres/) (probably more) pipeline by default for concurrent queries to avoid multiple network requests.

**Repro**

This can be reproduced using a simple pipelined psql (≥17) script:

```
\startpipeline
SELECT 1 AS a \bind \sendpipeline
\syncpipeline
SELECT pg_sleep(1) \bind \sendpipeline
\syncpipeline
SELECT 2 AS a \bind \sendpipeline
\syncpipeline
\endpipeline
```

A `postgres.js` repro which basically sends concurrent SQL queries (this is where I first caught it):

```javascript
import postgres from 'postgres'

const sql = postgres('postgresql://postgres.TENANT:PASS@HOST:6543/postgres', { max: 3, prepare: true })

const rows = await Promise.all(Array.from({ length: 20 }, (_, i) => sql`SELECT ${i}::int AS n`))
console.log('got', rows.length, 'results')
```

Snehil-Shah:

Avoids prematurely releasing the backend on the first-found `ReadyForQuery` message. 

The client hangs waiting for the rest of the replies, but Supavisor already releases the backend early, causing the remaining replies to be lost.

## What is the new behavior?

The client handler now communicates the expected `ReadyForQuery` messages to wait for, so we only release the backend when we are done with the batch.

## Additional context

Resolves #1061 

I spent good time on the tests. The integration tests correctly catch the regression (fails on main, passes here).

### DIFF
--- a/lib/supavisor/client_handler/protocol_helpers.ex
+++ b/lib/supavisor/client_handler/protocol_helpers.ex
@@ -97,11 +97,12 @@ defmodule Supavisor.ClientHandler.ProtocolHelpers do
         :transaction,
         %{tenant_feature_flags: tenant_feature_flags} = data
       ) do
-    if FeatureFlag.enabled?(tenant_feature_flags, "named_prepared_statements") do
-      MessageStreamer.handle_packets(data.stream_state, bin)
-    else
-      {:ok, data.stream_state, bin}
-    end
+    translate? = FeatureFlag.enabled?(tenant_feature_flags, "named_prepared_statements")
+
+    stream_state =
+      MessageStreamer.update_state(data.stream_state, &%{&1 | translate?: translate?})
+
+    MessageStreamer.handle_packets(stream_state, bin)
   end
 
   def process_client_packets(bin, _mode, data) do
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -45,6 +45,7 @@ defmodule Supavisor.ClientHandler do
   }
 
   alias Supavisor.Protocol.{FrontendMessageHandler, MessageStreamer}
+  require MessageStreamer
 
   alias Supavisor.Errors.{
     CheckoutTimeoutError,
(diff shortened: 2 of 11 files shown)


Snehil-Shah:
@v0idpwn I ran pgbench locally on my laptop (M2 8 cores, 16GB RAM).

**Setup**: `SELECT 1;`, 60 concurrent clients across 10 threads, for 15 seconds. I ran this setup 5 times to get a fair median.

I used a light query so we can focus on the actual code path overhead. The setup runs with `named_prepared_statements: false`, so we are purely testing the new framed path vs raw forwarding on main (before fix).

Here are the numbers:

| | median TPS | median latency | TPS range (5 rounds) |
|---|---:|---:|---|
| `main` (https://github.com/supabase/supavisor/tree/3dca3aa0d853ae309ee9941628dbc32a4c87a2ad) | 12,236 | 4.90 ms | 12,135 – 12,256 |
| this PR | 11,918 | 5.03 ms | 11,883 – 12,077 |
| **delta** | **−2.6 %** | **+2.7 %** | non-overlapping |

WDYT? I couldn't do a proper load test, although I feel this is enough for the story.

Snehil-Shah:
@v0idpwn Another thing I wanted to flag is that this PR changes the state shape of `DbHandler`, `ClientHandler`, and `BackendMessageHandler`. I presume this will need to be handled during a relup. I think we can write a hot upgrade migration for this using `:sys.replace_state` to handle old states. WDYT? Thought I would consult you first on how to approach this.

lib/supavisor/client_handler/protocol_helpers.ex
```diff
@@ -97,11 +97,12 @@ defmodule Supavisor.ClientHandler.ProtocolHelpers do
         :transaction,
         %{tenant_feature_flags: tenant_feature_flags} = data
       ) do
-    if FeatureFlag.enabled?(tenant_feature_flags, "named_prepared_statements") do
-      MessageStreamer.handle_packets(data.stream_state, bin)
-    else
-      {:ok, data.stream_state, bin}
-    end
+    translate? = FeatureFlag.enabled?(tenant_feature_flags, "named_prepared_statements")
```
v0idpwn, line 100: ~~Let's take the opportunity to drop this feature flag?~~

Edit: I'll do this separately with a more paced approach (default to on)

v0idpwn (commented):
This looks great to me! I'll run tests/benchmarks too on it and follow up if anything is needed.

v0idpwn:
Regarding the upgrade, we are going to merge this to main, which is targeting the v2.10 release, which is not out yet. 

We don't intend to relup from v2.9 to 2.10, so we don't need to make this relup compatible.

Snehil-Shah:
Not sure why the new js test keeps timing out, seems to run fine locally🤔

Snehil-Shah:
@v0idpwn Increased the timeout and decreased the iterations. Can we trigger CI to see if it fixes anything?

v0idpwn (approved):

PR #1080, opened 2026-07-14
feat(monitoring): add system network counter metrics via procfs

terrywilcox:
Polls /proc/net/netstat on each tick and emits ListenDrops and ListenOverflows as Prometheus gauges. These kernel counters surface OS-level connection drops that are invisible to application metrics.

No-ops silently on non-Linux systems.

Add new metrics for netstat

No metrics for netstat

## What is the new behavior?

new metrics

## Additional context

### DIFF
--- a/lib/supavisor/monitoring/net_stat.ex
+++ b/lib/supavisor/monitoring/net_stat.ex
@@ -1,0 +1,90 @@
+defmodule Supavisor.PromEx.Plugins.NetStat do
+  @moduledoc """
+  Polls system-wide network counters via procfs.
+  """
+
+  use PromEx.Plugin
+
+  @event_net_stat [:supavisor, :prom_ex, :osmon, :net_stat]
+  @prefix [:supavisor, :prom_ex]
+  @proc_net_netstat "/proc/net/netstat"
+
+  @impl true
+  def polling_metrics(opts) do
+    poll_rate = Keyword.get(opts, :poll_rate)
+
+    [
+      net_stat_metrics(poll_rate)
+    ]
+  end
+
+  defp net_stat_metrics(poll_rate) do
+    Polling.build(
+      :supavisor_osmon_net_stat_events,
+      poll_rate,
+      {__MODULE__, :execute_net_stat_metrics, []},
+      [
+        last_value(
+          @prefix ++ [:osmon, :net, :listen_drops],
+          event_name: @event_net_stat,
+          description: "Cumulative number of connections dropped due to a full accept queue.",
+          measurement: :listen_drops,
+          reporter_options: [prometheus_type: "counter"]
+        ),
+        last_value(
+          @prefix ++ [:osmon, :net, :listen_overflows],
+          event_name: @event_net_stat,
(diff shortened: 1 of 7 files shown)


lib/supavisor/monitoring/osmon.ex
```diff
@@ -191,4 +201,59 @@
        }}
     end
   end
+
+  @spec net_stat(Path.t()) :: {:ok, %{listen_drops: non_neg_integer(), listen_overflows: non_neg_integer()}} | :error
+  def net_stat(path \\ @proc_net_netstat) do
+    with {:ok, content} <- File.read(path),
```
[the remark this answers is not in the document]
terrywilcox: path defaults to the hardcoded /proc/net/netstat constant and is only overridden by tests with a fixture file, so it's not attacker-controlled.

lib/supavisor/monitoring/osmon.ex
```diff
@@ -191,4 +201,61 @@ defmodule Supavisor.PromEx.Plugins.OsMon do
...
+    end) || :error
+  end
+
+  defp net_stat_metrics(poll_rate) do
+    Polling.build(
+      :supavisor_osmon_net_stat_events,
+      poll_rate,
+      {__MODULE__, :execute_net_stat_metrics, []},
+      [
+        last_value(
```
v0idpwn: I think these should be counters, as they are cumulative

terrywilcox: I concur. I'll change it.

lib/supavisor/monitoring/osmon.ex
```diff
@@ -11,14 +11,17 @@ defmodule Supavisor.PromEx.Plugins.OsMon do
   @event_cpu_util [:prom_ex, :plugin, :osmon, :cpu_util]
   @event_cpu_la [:prom_ex, :plugin, :osmon, :cpu_avg1]
   @event_disk [:prom_ex, :plugin, :osmon, :disk]
+  @event_net_stat [:supavisor, :prom_ex, :osmon, :net_stat]
```
v0idpwn: I wonder if these should be in a separate module/prefix

terrywilcox: Yes, if we're planning on adding any more TCP metrics, a separate module would be a better idea. Will do.

v0idpwn (approved):

lib/supavisor/monitoring/net_stat.ex
```diff
@@ -0,0 +1,90 @@
...
+
+  # sobelow_skip ["Traversal.FileModule"]
+  @spec net_stat(Path.t()) ::
+          {:ok, %{listen_drops: non_neg_integer(), listen_overflows: non_neg_integer()}} | :error
+  def net_stat(path \\ @proc_net_netstat) do
+    with {:ok, content} <- File.read(path),
+         {:ok, stats} <- parse_net_stat(content) do
+      {:ok, stats}
+    else
+      _ -> :error
```
mentels, line 64: I find it useful to log if that fails

PR #1083, opened 2026-07-15
chore: rollback euw21 to v2.9.1

mentels:
Standby rollback for #1082.

**Do NOT merge unless rolling back the euw21 upgrade.**


mentels:
not needed

PR #1085, opened 2026-07-15
chore: rollback apne11 to v2.9.0

mentels:
Standby rollback for #1082.

**Do NOT merge unless rolling back the apne11 upgrade.**


mentels:
not needed

PR #1086, opened 2026-07-15
chore(supavisor-staging): update ap-southeast-1-blu to v2.9.10 

shamanime:
Platform: https://github.com/supabase/platform/pull/35743


mentels (approved):
FYI, we can achieve the same with https://github.com/supabase/supavisor/actions/workflows/stage.yml

But to mimic the production workflow that's better.

PR #1089, opened 2026-07-16
chore: upgrade to v2.9.10 in euw31

mentels:
platform: https://github.com/supabase/platform/pull/35768

PR #1091, opened 2026-07-16
chore: rollback euw31 to v2.9.2

mentels:
Standby rollback for #1089.

**Do NOT merge unless rolling back the euw31 upgrade.**


mentels:
Superseded by #1094 (corrected rollback target: 2.9.0 not 2.9.2)

PR #1113, opened 2026-07-22
feat: detect session-leak in transaction mode

v0idpwn:
Adds a new txn_mode_leak_action tenant option: ignore (default), log a warning, or return an error to the client.

Flags statements leaving state on the backend: session SET, set_config(_, _, false), DISCARD, session advisory locks, LISTEN/UNLISTEN, WITH HOLD cursors, temp tables, LOAD. 

Transaction-scoped variants (SET LOCAL, SET TRANSACTION, pg_advisory_xact_lock, ON COMMIT DROP, set_config(_, _, true)) are always  allowed.

Checked on Simple Query and Parse, across all statements in a query; first leak found is reported.

### DIFF
--- a/native/pgparser/src/lib.rs
+++ b/native/pgparser/src/lib.rs
@@ -10,4 +10,22 @@ fn statement_types(query: &str) -> Result<Vec<String>, String> {
     Ok(message)
 }
 
+#[rustler::nif]
+fn has_session_set(query: &str) -> Result<bool, String> {
+    let result = pg_query::parse(query).map_err(|_| "Error parsing query")?;
+
+    let found = result.protobuf.stmts.iter().any(|raw_stmt| {
+        match raw_stmt.stmt.as_deref().and_then(|node| node.node.as_ref()) {
+            Some(pg_query::NodeEnum::VariableSetStmt(stmt)) => {
+                // SET LOCAL, SET TRANSACTION and SET TRANSACTION SNAPSHOT are
+                // transaction-scoped, so they are safe in transaction mode
+                !stmt.is_local && stmt.name != "TRANSACTION" && stmt.name != "TRANSACTION SNAPSHOT"
+            }
+            _ => false,
+        }
+    });
+
+    Ok(found)
+}
+
 rustler::init!("Elixir.Supavisor.PgParser");
--- a/priv/repo/migrations/20260722000000_add_txn_mode_set_action.exs
+++ b/priv/repo/migrations/20260722000000_add_txn_mode_set_action.exs
@@ -1,0 +1,18 @@
+defmodule Supavisor.Repo.Migrations.AddTxnModeSetAction do
+  use Ecto.Migration
+
+  def change do
+    alter table("tenants", prefix: "_supavisor") do
+      add(:txn_mode_set_action, :string, null: false, default: "ignore")
+    end
+
+    create(
+      constraint(
+        "tenants",
+        :txn_mode_set_action_values,
+        check: "txn_mode_set_action IN ('ignore', 'log', 'error')",
+        prefix: "_supavisor"
+      )
+    )
+  end
+end
--- a/priv/repo/migrations/20260722000000_add_txn_mode_set_action.exs
+++ b/priv/repo/migrations/20260722000000_add_txn_mode_set_action.exs
@@ -11,7 +11,8 @@ defmodule Supavisor.Repo.Migrations.AddTxnModeSetAction do
         "tenants",
         :txn_mode_set_action_values,
         check: "txn_mode_set_action IN ('ignore', 'log', 'error')",
-        prefix: "_supavisor"
+        prefix: "_supavisor",
+        validate: false
       )
     )
   end
--- a/test/supavisor/protocol/frontend_message_handler_test.exs
+++ b/test/supavisor/protocol/frontend_message_handler_test.exs
@@ -1,6 +1,7 @@
 defmodule Supavisor.Protocol.FrontendMessageHandlerTest do
   use ExUnit.Case, async: true
 
+  import ExUnit.CaptureLog
   import Supavisor.Asserts
 
   alias Supavisor.Protocol.FrontendMessageHandler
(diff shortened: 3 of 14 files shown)


v0idpwn:
cc @Snehil-Shah figured you might want to take a look on this one, appreciate any opinions

native/pgparser/src/lib.rs
```diff
@@ -10,4 +10,22 @@ fn statement_types(query: &str) -> Result<Vec<String>, String> {
...
+                // SET LOCAL, SET TRANSACTION and SET TRANSACTION SNAPSHOT are
+                // transaction-scoped, so they are safe in transaction mode
+                !stmt.is_local && stmt.name != "TRANSACTION" && stmt.name != "TRANSACTION SNAPSHOT"
+            }
+            _ => false,
+        }
+    });
+
+    Ok(found)
+}
```
Snehil-Shah, line 29: @v0idpwn I did some AST parsing too in #1012. My main design approach was to let the rust layer expose basic primitives for interacting with pgparser (like returning the AST in JSON), instead of trying to touch application-specific logic. This way, elixir-land can own the application logic via happy JSON returned by the AST primitives it will call via NIF.

Here's a comment that gives some more context on how I wrote it: https://github.com/supabase/supavisor/pull/1012#discussion_r3301442671. 

This is mainly a discussion on the code structure and separation of concerns. Although I do feel writing the entire parsing logic in native code would be a performance win. WDYT on either of the approaches?

Snehil-Shah (commented):
@v0idpwn The idea feels pretty novel, I like it :). Just curious, is there any reference inspiration for this?

native/pgparser/src/lib.rs
Snehil-Shah: Also, I think we can extend this to include `LISTEN`/`UNLISTEN` and any other session-dependent queries, and nitpickingly update the function name to something like `is_session_scoped`?

v0idpwn: I like the idea of getting more things 🤔

v0idpwn: Regarding the parsing, my gut feeling is that we should reduce the data volume leaving the NIF as much as possible, unless outputting more allows us to perform less work somehow (e.g.: caching). But I haven't really reviewed this code in #1012 yet, and it might change my mind.

v0idpwn:
>Just curious, is there any reference inspiration for this?

If any other pooler implements it, I'm not aware. But this is a known footgun of transaction pooling and we regularly get reports of "pool poisoning" by our users, hence we want to improve the experience :)

Snehil-Shah:
@v0idpwn I'm aware of the footgun, but I haven't seen anything more than a documented warning. This will definitely be great UX.

native/pgparser/src/lib.rs
Snehil-Shah: @v0idpwn It can definitely reduce the work moving forward as we only parse the AST once via NIF, and do many things with it in elixir-land, from SET statement detection, to read-safe query detection for replica routing (and any other features that might require working with the AST). I don't have a concrete vision yet on how it might look like, just thinking out loud.

v0idpwn: Yeah... This makes sense. I wonder if we should output rust resources and act on them via nifs though, to avoid the type marshalling...

Snehil-Shah: @v0idpwn Agreed! Passing around rust resource refs seems like the right way to go. Looking back, I didn't think through performance as much in #1012 😅. Will update it once and if we define the template for doing this in this PR.

priv/repo/migrations/20260722000000_add_txn_mode_set_action.exs
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    create(
+      constraint(
+        "tenants",
+        :txn_mode_set_action_values,
+        check: "txn_mode_set_action IN ('ignore', 'log', 'error')",
+        prefix: "_supavisor"
+      )
+    )
```
mentels, line 17: Have you checked that it will NOT perform a full table scan? Or it's safe given our tables size?

Usually I'd got with a `NOT VALID` constraint but have no past experience wrt our `tenants` tables in prod.

test/supavisor/protocol/frontend_message_handler_test.exs
```diff
@@ -107,6 +108,138 @@ defmodule Supavisor.Protocol.FrontendMessageHandlerTest do
...
+      stream_state = with_set_statements_action(stream_state, :log)
+      bin = simple_query("SET statement_timeout = '1s'")
+
+      log =
+        capture_log(fn ->
+          assert {:ok, _, [^bin]} = MessageStreamer.handle_packets(stream_state, bin)
+        end)
+
+      assert log =~
+               "received session-level SET statement in transaction mode: SET statement_timeout = '1s'"
```
mentels, line 145: Shall we check if we have the `project` metadata on these logs?

test/supavisor/protocol/frontend_message_handler_test.exs
```diff
@@ -107,6 +108,138 @@ defmodule Supavisor.Protocol.FrontendMessageHandlerTest do
...
+    end
+
+    test "error action returns an error on Parse message", %{stream_state: stream_state} do
+      stream_state = with_set_statements_action(stream_state, :error)
+      bin = parse_message("SET statement_timeout = '1s'")
+
+      assert {:error, %Supavisor.Errors.SetStatementNotAllowedError{}} =
+               error = MessageStreamer.handle_packets(stream_state, bin)
+
+      assert_valid_error(error)
```
mentels, line 197: Will we see this error on the logs?

mentels (commented):
Looks great!

Just a few questions.

priv/repo/migrations/20260722000000_add_txn_mode_set_action.exs
v0idpwn: Yep, it performs a FTS, and its going to hold a lock in the table for the duration.

v0idpwn: Per the discussion, I have made it NOT VALID, and it will be made valid in a separate migration

test/supavisor/protocol/frontend_message_handler_test.exs
v0idpwn: yes, but notice its raised as a warning

test/supavisor/protocol/frontend_message_handler_test.exs
v0idpwn: when its logged here it doesn't include the project, but the project comes from the ClientHandler metadata.

config/runtime.exs
```diff
@@ -267,7 +267,9 @@ if config_env() != :test do
   config :supavisor, Supavisor.FeatureFlag, %{
     "named_prepared_statements" =>
       Supavisor.Helpers.get_env_bool("NAMED_PREPARED_STATEMENTS_ENABLED", false),
-    "app_name_metric" => Supavisor.Helpers.get_env_bool("APP_NAME_METRIC_ENABLED", false)
+    "app_name_metric" => Supavisor.Helpers.get_env_bool("APP_NAME_METRIC_ENABLED", false),
+    "check_simple_query_prepare" =>
+      Supavisor.Helpers.get_env_bool("CHECK_SIMPLE_QUERY_PREPARE_ENABLED", false)
```
v0idpwn, line 272: I added this because I think this check is low value and adds cost, so I'm disabling it by default.

Not exactly related to this PR, but adding the toggle was necessary for benchmarking all cases :)

lib/supavisor/protocol/simple_query_handler.ex
```diff
@@ -1,14 +1,25 @@
 defmodule Supavisor.Protocol.SimpleQueryHandler do
   @moduledoc """
-  Handles PostgreSQL Simple Query (Q) messages.
+  Restrictions on PostgreSQL Simple Query (Q) messages in transaction mode.
```
v0idpwn:
```suggestion
  Handles PostgreSQL Simple Query (Q) messages.
```

native/pgparser/src/lib.rs
```diff
@@ -1,13 +1,231 @@
...
+    if name == SET_CONFIG {
+        // set_config(setting, value, is_local); only the session-scoped form
+        // poisons the connection. A non-literal is_local (a parameter, say)
+        // could be either, and flagging it would reject calls that are in fact
+        // transaction-local, so it is let through
+        let session_scoped = func_call
+            .args
+            .get(2)
+            .and_then(bool_const_value)
+            .is_some_and(|is_local| !is_local);
```
v0idpwn, line 189: Potential follow up,  a bit hard: when dealing with the extended protocol, we don't have the parameters for `set_config` at the time we see it (handling of the PARSE packet). So we don't know if its `local = true` or `local = false`. To improve this, we'd need to somehow check at `Bind` time what was the parameter given.

lib/supavisor/protocol/session_leaks.ex
```diff
@@ -0,0 +1,63 @@
...
+  """
+  @spec describe(PgParser.session_leak()) :: String.t()
+  def describe(:session_set), do: "session-level SET statement"
+  def describe(:set_config), do: "session-level set_config() call"
+  def describe(:discard), do: "DISCARD statement"
+  def describe(:advisory_lock), do: "session-level advisory lock"
+  def describe(:listen), do: "LISTEN statement"
+  def describe(:hold_cursor), do: "WITH HOLD cursor"
+  def describe(:temp_table), do: "temporary table creation"
+  def describe(:set_constraints), do: "session-level SET CONSTRAINTS statement"
```
Snehil-Shah: Correct me if I'm missing something, but `SET CONSTRAINTS` is transaction-scoped and can't cause any session-level leaks: [ref](https://www.postgresql.org/docs/current/sql-set-constraints.html).

native/pgparser/src/lib.rs
```diff
@@ -1,13 +1,231 @@
...
+            let temp = stmt.into.as_ref().is_some_and(|into| {
+                into.rel
+                    .as_ref()
+                    .is_some_and(|rel| is_temp_persistence(&rel.relpersistence))
+                    && into.on_commit != ONCOMMIT_DROP
+            });
+
+            temp.then(atoms::temp_table)
+        }
+
```
Snehil-Shah, line 142: We are missing the `SELECT … INTO TEMP` case for temp tables, worth adding:

```suggestion

        // SELECT ... INTO [TEMP] t parses as a SelectStmt carrying an IntoClause
        pg_query::NodeEnum::SelectStmt(stmt) => {
            let temp = stmt.into_clause.as_ref().is_some_and(|into| {
                into.rel
                    .as_ref()
                    .is_some_and(|rel| is_temp_persistence(&rel.relpersistence))
                    && into.on_commit != ONCOMMIT_DROP
            });

            if temp {
                Some(atoms::temp_table())
            } else {
                func_call_leak(node)
            }
        }

```

Snehil-Shah (commented):
Looks great :)

native/pgparser/src/lib.rs
v0idpwn: Done it on this PR.

lib/supavisor/protocol/session_leaks.ex
v0idpwn: NIce catch!

native/pgparser/src/lib.rs
v0idpwn: good call, done

PR #1126, opened 2026-07-23
chore: upgrade to v2.9.10 in ap-northeast-2

shamanime:
Platform https://github.com/supabase/platform/pull/36110

Ref POOLER-402

### DIFF
--- a/deploy/upgrade/prod.config
+++ b/deploy/upgrade/prod.config
@@ -1,10 +1,10 @@
 [
-    {upgrade_from, "2.9.6"},
+    {upgrade_from, "2.9.7"},
     {upgrade_previous, "2.9.4"},
     {upgrade_to, "2.9.10"},
     % list() | [<<"all">>]
     {regions, [
-        <<"ca-central-1">>
+        <<"ap-northeast-2">>
             ]},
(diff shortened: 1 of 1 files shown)


deploy/upgrade/prod.config
```diff
@@ -1,10 +1,10 @@
 [
-    {upgrade_from, "2.9.6"},
+    {upgrade_from, "2.9.7"},
     {upgrade_previous, "2.9.4"},
```
deepthi, line 3: Do we use this for anything? It's not being kept in sync, Ideally it should be 2.9.6 as part of this PR.

shamanime: It's not necessary because it's a hard deploy but it goes on the name of the generated artifact so this matches the current state.

shamanime: No objections on reverting it though.

deepthi: I meant `upgrade_previous` should be 2.9.6 if we have any type of rollback mechanism. Will approve with this comment and we can revisit how `upgrade_previous` is being used.

deepthi (approved):

PR #1129, opened 2026-07-27
Chore/pooler 544 rollback to 2.9.6 in us west 1 sec

terrywilcox:
Rollback upgrade from v2.9.10 to v2.9.6 on us-west-1-sec
 
upgrade https://github.com/supabase/supavisor/pull/1128

Do not merge unless we need to rollback.


terrywilcox:
not needed

PR #1136, opened 2026-08-03
feat: parser optimization

dapeng1162:
# SQL Parser Rewrite and Per-Scheduler Parse Cache

This document describes the rework of `Supavisor.PgParser` shipped in this
PR: a C NIF linked directly against libpg_query 17 (replacing the
Rustler/`pg_query.rs` NIF), plus a per-scheduler cache of classification
results keyed by statement *shape*. It covers the motivation, the design
with emphasis on how correctness is guaranteed, and the measured results.

## 1. Background

Every query that arrives over the PostgreSQL simple protocol is classified
before routing: `Supavisor.Protocol.SimpleQueryHandler` calls
`Supavisor.PgParser.statement_types/1` on the query text to learn the
top-level statement kinds (`["SelectStmt"]`, `["InsertStmt"]`, ...). This
runs **on every client SQL string**, so classification is one of the
hottest code paths in the data plane.

The previous implementation had two problems:

1. **A full parse per query.** The result depends only on the statement
   *shape*, yet `SELECT * FROM t WHERE id = 1` and the identical query
   with `id = 2` were each parsed from scratch.
2. **An extra, expensive hop in the chain.** The NIF was a Rustler crate
   using `pg_query.rs` (6.1.0) — and `pg_query.rs` itself is only a thin
   Rust wrapper around libpg_query: it vendors the same libpg_query C
   sources into a Rust crate and exposes them through its own API. The
   real parser was always C, so this middle layer added nothing over the
   C library — only overhead. Concretely, the old path parsed through
   `pg_query_parse_protobuf`: libpg_query builds a complete **protobuf
   AST**, `pg_query.rs` decodes the entire tree into Rust structs — and
   then `statement_types()` read only the top-level statement kinds and
   returned a `Vec<String>`. The protobuf construction and the full-tree
   decode were pure overhead for this use case, and the Rust crate added
   a second toolchain to every build (release Dockerfiles, CI,
   hot-upgrade pipelines). The NIF can link libpg_query directly
   instead, which is exactly what this rework does.

Measured baseline (Linux aarch64, 10 cores, Elixir 1.18.2 / OTP 27.2,
`mix run --no-start bench/pg_parser.exs`, 2026-07-30): 4.35 µs for a
trivial insert, 100.87 µs for a 777-byte analytical SELECT, 176.85 µs for
a ~1.2 KB long IN list.

### 2.1 High level design

The rework changes how a simple-protocol query gets classified, in two
independently useful steps:

1. **Parse in C instead of Rust + protobuf.** The Rustler/`pg_query.rs`
   NIF is replaced by a plain C NIF (`nif/pgparser/`) linked directly
   against libpg_query 17: `pg_query_raw_parse` builds the raw parse tree
   and the NIF walks it (`List` of `RawStmt`) in C, returning only the
   top-level statement kind names. No protobuf AST is built, no tree
   crosses the NIF boundary, and the Rust toolchain disappears from the
   build.
2. **Cache classifications by statement shape.** The classification
   result depends only on a query's *shape*, so each scheduler keeps an
   LRU cache from a canonical encoding of the query's token stream to the
   statement kind list. On a hit the query is never parsed at all.

Both steps preserve the `Supavisor.PgParser.statement_types/1` contract —
`{:ok, ["SelectStmt", ...]} | {:error, "Error parsing query"}` —
byte-for-byte, so callers see no difference.

Per-query flow:

```
simple query
    │
    ▼
scan once with the PostgreSQL core scanner
    │
    ├─ not eligible (non-DML, multi-statement, scan error)
    │    └─► full C parse ──► result                         [bypass]
    │
    └─ eligible ──► per-scheduler LRU lookup
                      ├─ hit  ──► cached result
                      └─ miss ──► full C parse ──► insert ──► result
```

Every path that does not serve from the cache behaves exactly as if the
cache did not exist.

### 2.2 Low level design

**C NIF** (`nif/pgparser/`, plain C built with make):

- libpg_query is pinned to `17-6.2.1` with sha256 verification of the
  tarball; `nif/pgparser/patches/` can carry local patches (currently
  none are needed).
- The NodeTag → name table is generated at build time from `nodetags.h`,
  keeping the protobuf-compatible naming (`SelectStmt`, `InsertStmt`,
  ...).
- Only the statement kind names cross the NIF boundary — no protobuf, no
  AST marshalling, no Rust toolchain.

**Shape encoding.** On the cache path the SQL is scanned once by the
PostgreSQL core scanner — the same scanner the parser itself uses — and
encoded into a canonical token stream:

- keywords → their token codes;
- identifiers and operators → normalized bytes;
- `$n` parameters → the parameter number;
- literal values → folded to their literal *type* tag (integer, float,
  string, bit-string remain distinct);
- comments and whitespace are skipped; trailing semicolons are ignored.

**Cache key and value.** The canonical stream is hashed with seeded
XXH64 (64-bit); the key is the digest plus the token count and the
serialized length, and carries a versioned context header (see 2.3). The
cached value is the statement kind list (~300 bytes).

**Placement and sizing.** Caches are **per scheduler** via `enif_tsd`:
each scheduler owns a lock-free LRU cache, so the hot path takes no
locks, executes no atomics, and touches no shared cache lines.
`PARSER_CACHE_SIZE` (default 1024, `0` disables) bounds each scheduler's
cache; total memory is roughly `size × ~300 bytes × schedulers`.

**Observability.** `cache_stats/0` exposes aggregated counters (hits,
misses, bypasses split by reason, inserts, evictions, entries, parses,
parse_errors); the two headline counters are exported as
`parser_cache_hits` / `parser_cache_misses` Prometheus metrics.

### 2.3 Correctness

The cache must never return a result the parser itself would not have
returned. Five mechanisms enforce this, in order:

1. **Conservative eligibility.** Only single-statement
   SELECT/INSERT/UPDATE/DELETE/MERGE queries are eligible. Everything
   else — CTEs (`WITH`), `PREPARE`/`EXECUTE`, `DO`, `SET`, `EXPLAIN`,
   multi-statement payloads (detected from a top-level `;` in the token
   stream), and anything the scanner rejects — **bypasses** the cache and
   always takes the full-parse path, with exactly the same results and
   error semantics as if no cache existed.
2. **Same key ⇒ same result, by construction.** The shape encoding only
   folds aspects that provably cannot change the top-level statement
   kinds: literal *values* (folded per literal type, so `id = 1` and
   `id = 'a'` do **not** share a key), whitespace, comments, and trailing
   semicolons. Keywords, identifiers, operators, and parameter numbers
   all participate in the key. The first keyword is verified against the
   allowlist from the token stream itself. One exception class is
   excluded from shaping entirely: literals whose *content* is validated
   only at parse time, where the core scanner cannot see the outcome. In
   PostgreSQL 17 these are exactly two — unicode literals (`U&'...'` /
   `U&"..."` bodies and any `UESCAPE` clause, validated by the parser's
   token-filter layer via `str_udeescape`/`check_uescapechar`) and
   `float(p)` precisions (validated by a grammar action on the folded
   integer value). Two queries can share a token shape yet differ in
   parse success (`UESCAPE '!'` vs `UESCAPE '0'`, `float(53)` vs
   `float(55)`), so both bypass the cache. The parse-time error surface
   of the pinned libpg_query (grammar actions + token filter) was
   audited exhaustively: every other parse-time check depends on
   keywords, identifiers, parameters, or structure, all of which join
   the shape.
3. **Every insert is gated by a real parse.** An entry is written only
   after a full raw parse of that very SQL confirms exactly one
   allowlisted statement; the cached value is exactly what that parse
   returned. Lookups compare the full key: the 64-bit XXH64 digest, the
   token count, and the serialized size. With a 64-bit digest the
   birthday-bound collision probability is ~2.7e-12 at 1e4 cached shapes
   per scheduler, and a collision requires identical token count and
   serialized size on top. The seed is folded from 32 bytes of random
   material generated at boot; note that seeded XXH64 is *not* a keyed
   MAC, so this is only best-effort resistance against crafted
   collisions — deliberate cache poisoning is explicitly out of the
   threat model (the worst case of a collision is a wrong
   statement-kind list for the colliding query, and inserts remain
   gated by a real parse).
4. **Versioned context header.** Every key includes a magic, a format
   version, and `PG_PARSE_INFO_RULES_VERSION`. Any change to the
   extraction/shape rules bumps the version; entries written under old
   rules simply miss after a hot upgrade and age out via LRU — no stale
   results, no explicit flush.
5. **ABI-checked state reuse on hot upgrade.** The NIF instance state
   opens with a magic and `PG_NIF_STATE_ABI_VERSION`. The upgrade
   callback reuses the old instance — TSD key, caches, and secret all
   survive a code reload — only when both match, i.e. the old memory
   provably has the layout the new code was compiled against. A foreign
   priv (e.g. the former Rustler library's, on the first C-NIF upgrade)
   or a layout change fails the check: the new library performs a fresh
   load and leaves the old state to the old library's own unload, which
   frees it using the layout it was allocated with. The caches are
   volatile by design, so dropping them across an ABI break needs no
   migration. Any change to a reused layout or signature
   (`pg_nif_priv_t`, `pg_cache_t`, `pg_cache_entry_t`, `pg_shape_key_t`,
   `pg_parse_info_t`, the `pg_cache_*` functions) must bump
   `PG_NIF_STATE_ABI_VERSION`. A hot *downgrade* across an ABI break is
   unsupported: pre-magic code cannot recognize the new layout.

### 2.4 Test

Testing backs each correctness layer:

- 103 C unit checks (`make -C nif/pgparser test`): shape equivalence
  classes (literal/case/whitespace/comment/semicolon variants share a
  key; different tables, columns, quoted-identifier case, or literal
  types do not), XXH64 known-answer vectors from upstream's official
  sanity table, streaming-consistency checks (bulk vs. 1-byte vs.
  irregular chunk feeds), LRU semantics, stats identities, and
  bypass eligibility for parse-time-validated literals (unicode
  literals, `float(p)` precisions).
- A differential corpus at the Elixir level
  (`test/support/parse_shape_corpus.ex`, driven from
  `test/supavisor/pg_parser_test.exs`): same-shape groups must return
  identical results and converge to one cache entry; distinct-shape
  pairs must never hit each other's entries; allowlisted-but-invalid SQL
  never enters the cache; bypass families never hit or insert; the
  capacity identity `inserts == entries + evictions` holds; 1000
  concurrent same-shape calls stay consistent. Ordering regressions pin
  the parse-time-validated literal carve-out: `UESCAPE '!'` followed by
  `UESCAPE '0'` (and vice versa) and `float(53)` followed by `float(55)`
  (and vice versa) re-validate on every call and never hit or insert,
  while `float` without a precision stays cacheable.
- The classification table test covers multi-statement payloads, unicode
  identifiers, `E'...'` strings, empty/comment-only input, and the
  contract error shape.

### Reproducible ECS microbenchmark (x86_64)

The final parser comparison was run on 2026-08-03 on one Alibaba Cloud
`ecs.c9i.2xlarge` ordinary on-demand instance (8 vCPU / 16 GiB, Intel
Xeon 6982P-C, Ubuntu 22.04 x86_64). The runtime was Elixir 1.18.2 on OTP
27.2 / ERTS 15.2 with the JIT and eight online schedulers.

The Rust baseline was measured first, followed by the optimized build on
the same instance. Both source trees used the exact same
`bench/pg_parser.exs` (SHA-256
`6daa12e027d4964f6a4fa7e4605d03d627a2db7f077fdb48d2518ee3a36d5524`).
Each configuration ran in three fresh VM invocations with a 2-second
warmup, 5-second measurement, and `parallel: 1`. Every value below is
the median of the three per-run Benchee `average` values, not a selected
best run.

The revisions were:

- Rust + protobuf baseline: `origin/main` at
  `d366ef0056393692e669d4883e47abe0723d8435`;
- optimized base: `0da16d95155f63239ae56f4db27b75f2c32c4519`;
- optimized uncommitted binary diff: SHA-256
  `20a7b2309c1f4fd974f9d0a3000f35f930ee0505a472c835a8f0ce18c1b10560`;
- untracked `c_src/xxh64.h`: SHA-256
  `2e812fa055c74919908238258ee66cfdf6782cd2b94f378ba9bef9c5398c3ca2`.

"C parse" sets `PARSER_CACHE_SIZE=0`, isolating the C rewrite from the
cache. "C default" leaves the variable unset (`1024` entries per
scheduler); eligible repeated workloads are cache hits after warmup,
while DO, multi-statement, and CTE inputs intentionally bypass the cache.

| workload | Rust + protobuf | C parse, no cache | parse speedup | C default | total speedup |
| --- | ---: | ---: | ---: | ---: | ---: |
| trivial insert | 5.65 µs | 0.89 µs | 6.3x | 0.62 µs | 9.1x |
| dollar-quoted DO block | 3.81 µs | 1.06 µs | 3.6x | 1.22 µs | 3.1x |
| multi-statement | 19.55 µs | 1.99 µs | 9.8x | 2.22 µs | 8.8x |
| MERGE | 46.71 µs | 4.79 µs | 9.8x | 2.97 µs | 15.7x |
| INSERT multi-row RETURNING | 56.30 µs | 5.96 µs | 9.4x | 3.51 µs | 16.0x |
| jsonb + arrays | 64.69 µs | 5.73 µs | 11.3x | 3.22 µs | 20.1x |
| UPDATE with subquery | 71.27 µs | 5.71 µs | 12.5x | 3.18 µs | 22.4x |
| analytical SELECT (joins + windows) | 117.33 µs | 9.91 µs | 11.8x | 5.48 µs | 21.4x |
| CTE chain | 119.12 µs | 10.16 µs | 11.7x | 10.32 µs | 11.5x |
| deeply nested subquery | 120.51 µs | 8.29 µs | 14.5x | 4.13 µs | 29.2x |
| long IN list (~1.2 KB) | 207.25 µs | 13.13 µs | 15.8x | 6.01 µs | 34.5x |

The C rewrite alone is 3.6–15.8x faster than Rust + protobuf across the
suite. With the default cache setting, the complete optimized path is
3.1–34.5x faster. Bypass workloads can be slightly slower than C parse
alone because they conservatively scan before falling through to the
full parse; the table includes that cost rather than reporting only
cache-friendly inputs.

The second phase of the same harness explicitly exercises cache paths:

| path / workload | Rust baseline | optimized, default cache | speedup |
| --- | ---: | ---: | ---: |
| hit: repeated same SQL | 9.53 µs | 0.60 µs | 15.9x |
| hit: varying literals, same shape | 9.63 µs | 0.65 µs | 14.8x |
| bypass: PREPARE | 4.50 µs | 0.98 µs | 4.6x |
| miss: unique shape each call | 9.77 µs | 1.63 µs | 6.0x |
| hit: analytical SELECT | 118.04 µs | 5.48 µs | 21.5x |
| hit: long IN list (~1.2 KB) | 206.85 µs | 6.03 µs | 34.3x |

The Rust implementation has no cache. The hit/miss labels in its column
describe the SQL workload supplied by the identical harness, not an
internal Rust cache path.

### End-to-end cluster benchmark (pgbench, x86_64)

Three-instance setup on Alibaba Cloud cn-wulanchabu (2026-08-03), all
`ecs.c9i` ordinary on-demand compute instances on the internal network:

- pgbench 14.23 client: 16 vCPU / 32 GiB (drives the load);
- Supavisor: 8 vCPU / 16 GiB, `MIX_ENV=prod`, eight online schedulers,
  tenant `bench` (`pool_size=100`, transaction mode,
  `NAMED_PREPARED_STATEMENTS_ENABLED=true` unless noted;
  `client_heartbeat_interval=0` for clean long runs);
- PostgreSQL 14.23: 32 vCPU / 64 GiB, pgbench scale 500 in 16 GiB
  `shared_buffers`. A direct-to-PG run with the same 1000 clients
  measured 592,012 TPS (5-second-window median; 583,487 TPS end-of-run),
  confirming roughly 6x backend headroom.

Each configuration restarts the Supavisor VM with the relevant build and
flags. Reported figures are the **median of pgbench's 5-second progress
windows over a 2-minute steady state** (`-P 5`): the end-of-run summary
can be distorted by a pre-existing mass-disconnect teardown issue that
reproduces identically on both builds and is unrelated to this PR. In
every parsing configuration the BEAM uses 7.90–7.94 of the eight
available cores, so throughput differences directly reflect the CPU
cost of the parse path. CPU and RSS are medians of the saturated
`pidstat` samples; proxy CPU is BEAM CPU normalized to eight vCPUs.

**pgbench select-only** (`pgbench -n -S -c 1000 -j 16`):

| configuration | TPS (median) | TPS range | avg latency | proxy CPU | BEAM cores | BEAM RSS |
| --- | --- | --- | --- | --- | --- | --- |
| A: master, Rust `pg_query.rs` parse | 80,158 | 72.0k–80.4k | 12.529 ms | 99.3% | 7.94 | 254 MB |
| B: this PR, cache hit (`PARSER_CACHE_SIZE=1024`) | 99,877 | 89.8k–100.6k | 10.049 ms | 98.8% | 7.90 | 258 MB |
| C: this PR, cache disabled (`PARSER_CACHE_SIZE=0`) | 98,531 | 93.1k–98.9k | 10.173 ms | 98.8% | 7.91 | 252 MB |

Reference rows with parsing skipped entirely
(`NAMED_PREPARED_STATEMENTS_ENABLED=false`): this branch 105,904 TPS,
master 104,713 TPS — within 1.1% of each other, confirming the two
builds are otherwise equivalent.

After the cached load-test series, parser counters showed
`hits=19,315,203`, `misses=parses=inserts=entries=8` (one entry per
scheduler), `evictions=0`, and `parse_errors=0`.

**Load test** (the README `VALUES (...)` select through the same
simple-query classification path, `pgbench -n -f values.sql`, 30 s per
concurrency level):

| clients | A: master, Rust parse (TPS) | B: this PR, cache hit (TPS) | A latency ms | B latency ms |
| --- | --- | --- | --- | --- |
| 1 | 6,729 | 7,520 | 0.149 | 0.133 |
| 2 | 13,237 | 15,199 | 0.151 | 0.132 |
| 4 | 21,398 | 26,224 | 0.187 | 0.152 |
| 8 | 35,388 | 41,814 | 0.226 | 0.191 |
| 16 | 57,288 | 69,274 | 0.279 | 0.231 |
| 32 | 67,553 | 90,373 | 0.474 | 0.354 |
| 64 | 72,992 | 100,685 | 0.877 | 0.636 |
| 128 | 75,832 | 104,874 | 1.688 | 1.220 |
| 256 | 75,076 | 103,255 | 3.406 | 2.479 |
| 512 | 74,056 | 101,980 | 6.911 | 5.019 |

Reading the two tables together:

- End-to-end, B beats A by **+24.6% TPS and −19.8% latency** on
  select-only, and holds a **~+38% higher plateau** (102–105k vs
  74–76k TPS) under the cheaper VALUES query, where the proxy's
  per-query parse cost is a larger share of the work. Latency at 512
  clients drops from 6.911 ms to 5.019 ms (−27.4%).
- The C NIF alone delivers the bulk of the win (C vs A: +22.9%); the
  shape cache adds 1.4% on top on an already CPU-saturated proxy.
- Framed as the cost of having a parse path at all: master's Rust path
  spends **23.4%** of end-to-end throughput (80.2k vs 104.7k unparsed);
  this PR's parse-plus-cache path spends **5.7%** (99.9k vs 105.9k).
- BEAM memory stays flat in all configurations (~252–258 MB RSS); the
  cache's per-scheduler bound (`1024 × ~300 B × 8 schedulers`) is
  negligible.

## 4. Conclusion

The large, unconditional win is the C rewrite: the reproducible ECS run
measured a 3.6–15.8x faster no-cache classification path, with the Rust
toolchain and the protobuf AST removed from both the hot path and the
build. The complete optimized path with the default cache was
3.1–34.5x faster across the suite, including conservative bypass costs.
The cluster benchmark shows this holding end to end: on a CPU-saturated
8-vCPU proxy, the Rust parse path costs 23.4% of pgbench select-only
throughput, while this PR's full parse-plus-cache path costs 5.7% — net
+24.6% TPS and −19.8% latency over the Rust baseline, and a ~38% higher
plateau under the README load test.

The cache is intentionally more modest in latency terms, and it is worth
stating this plainly: **a cache hit is about 1.4–2.2x faster than a plain
C parse** in the reproducible ECS run (0.62 vs 0.89 µs trivial; 5.48 vs
9.91 µs analytical; 6.01 vs 13.13 µs long IN list). The scanner pass is
a shared, irreducible per-byte cost of both the hit and the miss path, so
hits cannot be orders of magnitude cheaper.

What the hit path does eliminate is **memory-subsystem work**, and that
is the point. A miss (or an uncached parse) pays a libpg_query memory
context setup and teardown, a full AST's worth of node allocations — and
on the old path, a protobuf tree on top — for every single query. A hit
performs one bounded scan, one incremental hash, and a ~2 ns LRU lookup,
then serves a pre-built ~300-byte result; no memory context, no AST, no
per-query transient allocation, and (because caches are per scheduler)
no locks and no cross-core cache-line traffic. Under extreme load — a
Supavisor cluster classifying every simple query from millions of client
connections — the benefit compounds through reduced allocator pressure
and fewer allocation-driven tail-latency spikes, rather than through the
mean-latency delta alone. For workloads where the cache is unwanted, it
can be disabled with `PARSER_CACHE_SIZE=0`, falling back to the already
3.6–15.8x faster plain C parse.


v0idpwn:
@dapeng1162 I appreciate the PR, but I'm not sure if this is something we are willing to maintain at the moment. I'm curious whether you are deploying Supavisor in conditions this change is needed. Could you please share a bit more about your use case? 

Thank you.

PR #1137, opened 2026-08-03
feat: add server_idle_timeout

Snehil-Shah:
Exposes the hardcoded poolboy idle_timeout as a new `server_idle_timeout` user config option. This mirrors the behavior of pgbouncer's `server_idle_timeout` config option. Above the minimum warm pool, each idle connections is closed once their timeout completes (preserves the existing 5 minutes as default).


Snehil-Shah:
cc @v0idpwn

PR #1139, opened 2026-08-10
fix: add project metadata to syn conflict resolution logs

terrywilcox:
Attach project/user/mode/db_name/type Logger metadata to resolve_registry_conflict/4 logs, matching the existing on_process_unregistered/5 convention, so pool conflict logs can be filtered by tenant.

POOLER-605

### DIFF
--- a/lib/supavisor/syn_handler.ex
+++ b/lib/supavisor/syn_handler.ex
@@ -29,12 +29,16 @@ defmodule Supavisor.SynHandler do
   @impl true
   def resolve_registry_conflict(
         :tenants,
-        id,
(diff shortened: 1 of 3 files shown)


lib/supavisor/syn_handler.ex
```diff
@@ -29,12 +29,16 @@ defmodule Supavisor.SynHandler do
   @impl true
   def resolve_registry_conflict(
         :tenants,
-        id,
+        Supavisor.id(type: type, tenant: tenant, user: user, mode: mode, db: db, search_path: _) =
```
v0idpwn:
```suggestion
        Supavisor.id(type: type, tenant: tenant, user: user, mode: mode, db: db) =
```

v0idpwn (approved):

PR #1144, opened 2026-08-11
chore: rollback eu-west-2-sec to v2.9.8

mentels:
Standby rollback for #1143.

**Do NOT merge unless rolling back the eu-west-2-sec upgrade.**


terrywilcox (approved):

mentels:
not needed

PR #1160, opened 2026-08-25
fix: bypass :inet_gethost_native for literal IP addresses

mentels:
inc-737

### DIFF
--- a/test/supavisor/helpers_test.exs
+++ b/test/supavisor/helpers_test.exs
@@ -72,19 +72,19 @@ defmodule Supavisor.HelpersTest do
   end
 
   describe "detect_ip_version/1" do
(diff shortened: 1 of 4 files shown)


terrywilcox (approved):

test/supavisor/helpers_test.exs
```diff
@@ -70,6 +70,24 @@ defmodule Supavisor.HelpersTest do
       assert new_min_heap_words >= expected_words
     end
   end
+
+  describe "detect_ip_version/1" do
+    test "IPv4 literal is detected without touching the resolver" do
```
v0idpwn: Nit: it's not validating it doesn't touch the resolver, so why claim that

PR #1169, opened 2026-08-29
ci: publish Docker Hub releases from release tags

avallete:
## Summary
- `Release on Dockerhub` only ran on pushes to `main` when `VERSION` changed. Version bumps now land on the release line (the `v2.9` branch), so Hub never got 2.9.8–2.9.11 (2.9.12 was published manually from a side branch). The old trigger was also a footgun: it path-filtered on the workflow file itself, so merging any change to it — including the previous version of this PR — would have rebuilt the stale 2.9.7 from main's much newer tree and overwritten the published Hub tag, then mirrored that to ECR/GHCR.
- Replace it with tag-anchored publishing:
  - `push` on `v*` tags publishes the tagged version automatically as releases happen. GitHub reads the workflow from the pushed tag, so this file must also be cherry-picked to the `v2.9` branch to cover future tags cut there — merging to main alone is not enough.
  - `workflow_dispatch` with a required `version` input backfills tags cut before the trigger existed. The dispatched ref is only used to locate the workflow file; the validated input decides what gets built.
- Every run checks out `refs/tags/v<version>` explicitly (the build jobs previously used the Git context of the triggering ref) and refuses to publish if the tag's `VERSION` file disagrees with the version being published.
- Optional `otp_version` / `debian_version` dispatch inputs override the Dockerfile `ARG`s for old tags whose pinned OTP 27.2.1 can no longer fetch from builds.hex.pm (`key_usage_mismatch`), so backfills don't need patched side branches.

## Test plan
- [x] actionlint passes (only pre-existing warning is the custom `arm-runner` label)
- [x] Version-derivation logic tested locally: tag push, dispatch input with/without leading `v`, build-arg emission, and rejection of non-version inputs (`main`, shell metacharacters)
- [ ] Merge this to main, then cherry-pick the same commit to `v2.9` so future tags auto-publish
- [ ] Backfill the missing releases: `gh workflow run publish_docker.yml --ref main -f version=2.9.8 -f otp_version=27.3.4 -f debian_version=bullseye-20260610-slim` (repeat for 2.9.9, 2.9.10, 2.9.11)
- [ ] Confirm `docker.io/supabase/supavisor:2.9.8`–`2.9.11` exist (amd64 + arm64) and are mirrored to ECR/GHCR
- [ ] Next release: push the `v2.9.13` tag and confirm the workflow runs with no manual dispatch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VvYtYTmAonGS6aw3rLQRfD

### DIFF
--- a/.github/workflows/publish_docker.yml
+++ b/.github/workflows/publish_docker.yml
@@ -1,27 +1,84 @@
 name: Release on Dockerhub
 
+# Version bumps land on the release line (e.g. the v2.9 branch), not main, so
+# the old "push to main when VERSION changes" trigger stopped publishing after
+# 2.9.7. Publishing is now driven by the release tag itself:
+#
+#   - Pushing a tag v<version> publishes supabase/supavisor:<version> built
+#     from that tag's tree. GitHub reads the workflow from the pushed tag, so
+#     this only fires for tags whose tree contains this file: it must live on
+#     each active release branch (e.g. v2.9), not just main.
+#   - Tags cut before this trigger existed can be backfilled with a manual
+#     dispatch from main: the version input decides what is built and
+#     published; the dispatched ref is only used to locate this file.
+#
+# Both paths check out refs/tags/v<version> and verify it against the VERSION
+# file at that tag, so a run can never rebuild a published version from the
+# wrong tree.
+
 on:
   push:
-    branches:
-      - main
-    paths:
-      - ".github/workflows/publish_docker.yml"
-      - "VERSION"
-  # VERSION bumps after 2.9.7 land on the release-tag line, not main, so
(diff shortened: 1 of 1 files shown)


avallete:
`supabase/supavisor:2.9.12` is on Docker Hub now (amd64 + arm64) from https://github.com/supabase/supavisor/actions/runs/33245742275 — dispatched on `avallete/publish-docker-v2.9.12` (v2.9.12 + OTP 27.3.4 so `mix local.hex` can talk to hex.pm).

This PR is still the lasting fix so the next tag does not need a side branch.

shamanime (approved):
Thank you 💚

Issue #1172, opened 2026-09-01
Transaction mode: a backend goes back to the pool with a transaction still open on it, and the reply is never relayed


crimble:
**Summary:** in transaction mode, Supavisor sometimes returns a
backend to its own pool with a transaction open on it and never relays the reply to the client that
issued it. The client waits for ever, Postgres shows the backend idle, and sixty seconds later
`idle_in_transaction_session_timeout` kills it.

## Symptom

A `begin read only` reaches Postgres, executes, and is answered. The client never receives the
reply. The connection is left checked out on the client side and idle in a transaction on the
server side, and Supavisor's own log then records:

```
DbHandler: Session terminated by server while idle in the pool: terminating connection due to idle-in-transaction timeout (25P03)
```

Four or five times per run. The same log carries `DbHandler: Session terminated by server while
checked out by a client`.

For a Next.js server with a fixed connection pool this is fatal rather than merely slow. Each
occurrence removes one connection permanently. With `max: 10`, ten occurrences leave every
authenticated request queueing behind a connection that will never answer. Measured on the wedged
server: 459 transactions waiting, 0 running, 10 connections holding, while the unauthenticated page
that touches no database served in 6 to 7ms throughout.

## Environment

| | |
| --- | --- |
| Pooler | Supabase transaction pooler, port 6543, region eu-west-2 |
| Client | postgres.js 3.4.9 |
| Runtime | Node v22.22.2 |
| Framework | Next.js 16.3.1 |
| Pool options | `max: 10`, `idle_timeout: 20`, `prepare: false` (the transaction pooler cannot hold a prepared statement across a transaction) |
| Server setting | `idle_in_transaction_session_timeout = 60s`, set on the application role |

## Reproduction

This does not need our repository. It needs a Node HTTP server in front of the transaction pooler
with a fixed pool, and a client that gives up mid-request.

1. Serve a page (production build, not dev) whose render opens one `read only` transaction on a
   pooled connection and runs about twenty statements inside it. Pool `max: 10`, `prepare: false`.
2. Issue **400 HTTP requests** to that page, **six in flight at a time**, **paced 40ms apart**, and
   **abort each one from the client** at a random point **between 5ms and 205ms** after it starts.
   All 400 aborted, nothing completed. The driver finishes in about 10 seconds.
3. Then make **one ordinary, un-aborted request** to the same page.

**What happens.** Step 3 never answers. Measured with the unmodified client (`sql.begin`) on
[...]

PR #1176, opened 2026-09-01
fix: remove stale Sync handler

Snehil-Shah:
Following up to #1079, I noticed a stale message handler in the client handler that directly sends any packet that starts with a `Sync` message to the backend without going through our "expected-rfq-counting" machinery. So any TCP packet that happens to start with `Sync` breaks our expected count and releases the backend earlier than required.


Snehil-Shah:
@v0idpwn I think this might be the root cause for #1172, due to symptoms of premature connection checkin.

v0idpwn (approved):
This looks right conceptually to me, nice catch! Could we perhaps get a failing test, or you think it'd be too hard?

Snehil-Shah:
@v0idpwn Tried to sketch one out using Claude. It catches the regression correctly and makes sense (fails on main, passes here)

test/integration/transaction_pipelining_test.exs
```diff
@@ -28,6 +28,31 @@ defmodule Supavisor.Integration.TransactionPipeliningTest do
...
       assert recv_ready_for_queries(sock, 1) == 1
     end
+
+    test "delivers every reply when a segment begins with a Sync (#{tenant})" do
+      sock = connect(unquote(tenant))
+
+      # A slow first statement keeps the ClientHandler :busy, so the second
+      # write is guaranteed to arrive mid-batch. That makes the seam
+      # deterministic instead of relying on where the kernel happens to split
+      # a coalesced pipelined write.
```
v0idpwn:
```suggestion
      # A slow first statement keeps the ClientHandler :busy when the Sync arrives
```

test/integration/transaction_pipelining_test.exs
```diff
@@ -28,6 +28,28 @@ defmodule Supavisor.Integration.TransactionPipeliningTest do
       :ok = :gen_tcp.send(sock, :pgo_protocol.encode_query_message("SELECT 1"))
       assert recv_ready_for_queries(sock, 1) == 1
     end
+
+    test "delivers every reply when a segment begins with a Sync (#{tenant})" do
```
v0idpwn:
```suggestion
    test "regression: delivers every reply when a segment begins with a Sync (#{tenant})" do
```

test/integration/transaction_pipelining_test.exs
```diff
@@ -28,6 +28,28 @@ defmodule Supavisor.Integration.TransactionPipeliningTest do
...
+
+      # A slow first statement keeps the ClientHandler :busy when the Sync arrives
+      :ok = :gen_tcp.send(sock, :pgo_protocol.encode_query_message("SELECT pg_sleep(0.3)"))
+      Process.sleep(50)
+
+      # The previous statement's trailing Sync, followed by a whole further
+      # statement: the shape a pipelined write takes when the segment boundary
+      # lands on an Execute|Sync seam. The second statement is slow too, so its
+      # reply lands in a later read from the backend, ie: after the premature
+      # release, which is when replies get dropped.
```
v0idpwn:
```suggestion
      # A sync, plus a query message.
      #
      # In the buggy implementation, the sync at the beginning of the packet would cause the 
      # following ReadyForQuery to trigger a checkin, and the query response wouldn't be received 
```

v0idpwn (approved):
Looks good, I just adjusted the comments

v0idpwn:
Sorry for the lag, took a few CI iterations :)

Anish-1205:
Well done @Snehil-Shah 👏👏👏

PR #1185, opened 2026-09-08
fix: replay parse before named statement describe

### LINKED ISSUE #1174: Supavisor transaction pooling does not replay `Parse` before a standalone `Describe`
## Supavisor version

v2.9.7 (main `49466b5`), transaction mode, `named_prepared_statements=true`.

## Description

After a named statement has been prepared, transaction pooling can route the next operation to a different backend. With pgJDBC 42.7.7, `PreparedStatement.getMetaData()` sends a standalone statement `Describe` and returns SQLSTATE `26000` (`prepared statement does not exist`), although `executeQuery()` on the same object succeeds immediately afterwards.

Supavisor translates a named `Describe` and forwards it without replaying the saved `Parse` ([`prepared_statements.ex`](https://github.com/supabase/supavisor/blob/49466b5/lib/supavisor/protocol/prepared_statements.ex#L187-L196), [`db_handler.ex`](https://github.com/supabase/supavisor/blob/49466b5/lib/supavisor/db_handler.ex#L1127-L1129)). Its `Bind` path replays the saved `Parse` when the selected backend lacks the statement ([`db_handler.ex`](https://github.com/supabase/supavisor/blob/49466b5/lib/supavisor/db_handler.ex#L1084-L1105)).

## Reproduce

1. Start PostgreSQL 16 and Supavisor in transaction mode with named prepared statements enabled and a pool with at least two backend connections.
2. Use PostgreSQL JDBC driver 42.7.7. Compile the program below and pass complete JDBC URLs for the direct and Supavisor endpoints. The program sets `prepareThreshold=1` so the first execution creates a server-side named statement.

```sh
javac -cp postgresql-42.7.7.jar reproduce.java
java -cp .:postgresql-42.7.7.jar Reproduce <direct-jdbc-url> <supavisor-jdbc-url>
```

```java
import java.sql.*;

class Reproduce {
  static Connection connect(String url) throws SQLException {
    return DriverManager.getConnection(url + (url.contains("?") ? "&" : "?")
        + "prepareThreshold=1&sslmode=disable");
  }

  static void run(String label, String url) throws Exception {
    try (Connection c = connect(url);
         PreparedStatement p = c.prepareStatement("SELECT ?::int")) {
      p.setInt(1, 7);
      try (ResultSet r = p.executeQuery()) { r.next(); }

      Thread holder = new Thread(() -> {
        try (Connection h = connect(url); Statement s = h.createStatement()) {
          s.execute("SELECT pg_sleep(2)");
        } catch (SQLException e) { throw new RuntimeException(e); }
      });
      holder.start();
      Thread.sleep(100);
      try {
        p.getMetaData();
        System.out.println(label + " describe: ok");
      } catch (SQLException e) {
        System.out.println(label + " describe: " + e.getSQLState() + " "
            + e.getMessage().split("\\R", 2)[0]);
      }
      try (ResultSet r = p.executeQuery()) {
        r.next();
        System.out.println(label + " execute: " + r.getInt(1));
      }
      holder.join();
    }
  }

  public static void main(String[] args) throws Exception {
    run("direct", args[0]);
    run("supavisor", args[1]);
  }
}
```

## Expected behavior

Both endpoints should print `describe: ok` and `execute: 7`.

## Actual behavior

The direct endpoint prints:

```text
direct describe: ok
direct execute: 7
```

Supavisor prints a `26000` error for `describe`, followed by a successful execution:

```text
supavisor describe: 26000 ERROR: prepared statement "sv_..." does not exist
supavisor execute: 7
```

dhruvxvaishnav:
## Summary

Named statement `Describe` messages were translated and forwarded without ensuring that the selected transaction-pool backend had seen the corresponding `Parse`. This caused clients such as pgJDBC to receive SQLSTATE `26000` from `PreparedStatement.getMetaData()` after a backend switch, even though executing the statement worked through the existing `Bind` replay path.

This change carries the cached `Parse` with named statement `Describe` packets and lets `Describe` use the same backend storage check and replay behavior as `Bind`. The replayed `ParseComplete` is intercepted so the client sees only the response to its original `Describe`.

Portal and unnamed statement describes remain unchanged.

Closes #1174

## Tests

- `mix test test/supavisor/protocol/prepared_statements_test.exs test/supavisor/protocol/debug_test.exs test/supavisor/db_handler_test.exs` (75 tests, 0 failures)
- `mix test test/integration/prepared_statements_test.exs` (9 tests, 0 failures)
- `mix format --check-formatted` on all changed files
- `mix dialyzer`
- Repeated the issue pgJDBC scenario with `prepareThreshold=1`: direct PostgreSQL and Supavisor both returned metadata successfully and executed with result `7`

The full `mix test` suite was also attempted in WSL on a Windows-mounted checkout. It completed 679 tests with 20 failures outside the changed areas, confined to an existing CRLF-sensitive assertion, 300 ms timing-sensitive tests, and peer-node startup exceeding its fixed 5 second timeout. The focused and prepared-statement suites pass cleanly; native Linux CI remains the authoritative full-suite result.

### DIFF
--- a/lib/supavisor/db_handler.ex
+++ b/lib/supavisor/db_handler.ex
@@ -1083,11 +1083,15 @@ defmodule Supavisor.DbHandler do
   defp take_chunk([], _remaining, acc), do: {:lists.reverse(acc), []}
 
   # If the prepared statement exists for us, it exists for the server, so we just send the
-  # bind to the socket. If it doesn't, we must send the parse pkt first.
+  # packet to the socket. If it doesn't, we must send the parse pkt first.
   #
-  # If we received a bind without a parse, we need to intercept the parse response, otherwise,
-  # the client will receive an unexpected message.
-  defp handle_prepared_statement_pkt({:bind_pkt, stmt_name, pkt, parse_pkt}, {iodata, data}) do
+  # If we replay a parse, we need to intercept the parse response, otherwise the client will
+  # receive an unexpected message.
+  defp handle_prepared_statement_pkt(
+         {packet_type, stmt_name, pkt, parse_pkt},
+         {iodata, data}
+       )
+       when packet_type in [:bind_pkt, :describe_pkt] do
     storage_mod = data.prepared_statements_storage
 
     if storage_mod.member?(data.prepared_statements, stmt_name) do
@@ -1128,10 +1132,6 @@ defmodule Supavisor.DbHandler do
              )
            end)
      }}
-  end
-
-  defp handle_prepared_statement_pkt({:describe_pkt, _stmt_name, pkt}, {iodata, data}) do
-    {[pkt | iodata], data}
   end
 
   # If we stop generating unique id per statement, and instead do deterministic ids,
--- a/lib/supavisor/protocol/debug.ex
+++ b/lib/supavisor/protocol/debug.ex
@@ -21,8 +21,8 @@ defmodule Supavisor.Protocol.Debug do
   @type message_source :: :frontend | :backend
   @type packet :: binary()
   @type structured_packet ::
-          {:parse_pkt | :close_pkt | :describe_pkt, String.t(), packet()}
-          | {:bind_pkt, String.t(), packet(), packet()}
+          {:parse_pkt | :close_pkt, String.t(), packet()}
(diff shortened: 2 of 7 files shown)


v0idpwn (approved):
Looks great to me!

PR #1186, opened 2026-09-08
fix: forward client_ip internal pooling

staaldraad:

Bug fix

When a pool already exists on Node A, Node B will forward the connection. With JIT auth, Node A would end up seeing the IP address of Node B, rather than the connecting external client's IP address. This IP address would be used in network restriction checks, leading to failed checks as the internal address is unexptected.

## What is the new behavior?

When forwarding a connection to an existing pool on another node, forward the client_ip of the external client (for JIT connections). This ensures JIT authentication uses the correct external IP address for both authz and logging.

Fixes authz for JIT seeing the internal ip address of the node forwarding the connection, rather than the external client's ip.

## Additional context

Fixes instances where JIT authentication would fail on some requests when using ip restrictions. This would occur when a load balancer is in the mix and it was non-deterministic which node would receive the connection.

### DIFF
--- a/lib/supavisor/client_handler/protocol_helpers.ex
+++ b/lib/supavisor/client_handler/protocol_helpers.ex
@@ -98,7 +98,8 @@ defmodule Supavisor.ClientHandler.ProtocolHelpers do
   """
   @spec effective_peer_ip(local? :: boolean(), forwarded_ip :: String.t() | nil, String.t()) ::
           String.t()
-  def effective_peer_ip(_local? = true, forwarded_ip, socket_peer_ip) when is_binary(forwarded_ip) do
+  def effective_peer_ip(_local? = true, forwarded_ip, socket_peer_ip)
+      when is_binary(forwarded_ip) do
     case :inet.parse_strict_address(to_charlist(forwarded_ip)) do
       {:ok, ip} -> List.to_string(:inet.ntoa(ip))
       {:error, _} -> socket_peer_ip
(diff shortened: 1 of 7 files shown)


v0idpwn (approved):

lib/supavisor/db_handler.ex
```diff
@@ -304,7 +305,11 @@ defmodule Supavisor.DbHandler do
             options = %{
               "search_path" => Supavisor.id(data.id, :search_path),
               "client_tls" => if(data.proxy, do: to_string(data.client_tls)),
-              "jit" => if(data.proxy, do: to_string(data.client_jit))
+              "jit" => if(data.proxy, do: to_string(data.client_jit)),
+              # original client's IP, so the pool node attributes the
+              # connection (JIT auth, circuit breaker, logs) to the client
+              # rather than to this node
```
v0idpwn:
```suggestion
```

lib/supavisor/client_handler/protocol_helpers.ex
```diff
@@ -78,12 +78,35 @@ defmodule Supavisor.ClientHandler.ProtocolHelpers do
...
+
+  Proxied connections arrive on a `local: true` listener from a peer node, so the
+  socket's peer is that node rather than the client. The forwarding node passes the
+  original client's IP in the `client_ip` startup option; we use it only when the
+  listener is local (never reachable by external clients) and the value is a valid
+  IP address. Otherwise the socket's peer IP is kept.
+  """
+  @spec effective_peer_ip(local? :: boolean(), forwarded_ip :: String.t() | nil, String.t()) ::
+          String.t()
+  def effective_peer_ip(true, forwarded_ip, socket_peer_ip) when is_binary(forwarded_ip) do
```
v0idpwn:
```suggestion
  def effective_peer_ip(_local? = true, forwarded_ip, socket_peer_ip) when is_binary(forwarded_ip) do
```

v0idpwn: Just makes it obvious in the code what the value means :)

PR #1187, opened 2026-09-08
feat: subscribe for a free client slot before rejecting with EMAXCONN

shamanime:
Instead of rejecting right away, subscribe and briefly wait for a connection slot to be assigned, if nothing is freed on time returns the expected EMAXCONN error.

Similar to #1179 but waits for the Manager to assign a slot instead of retrying after a delay.

### DIFF
--- a/lib/supavisor/client_handler.ex
+++ b/lib/supavisor/client_handler.ex
@@ -1050,9 +1050,10 @@ defmodule Supavisor.ClientHandler do
   # Completes connection setup once this client holds a pool slot, whether it was granted
   # synchronously by `Supavisor.subscribe/1` or handed over later by the Manager.
   defp finish_subscribe(data, opts) do
-    with manager_ref = Process.monitor(opts.workers.manager),
-         data = Map.merge(data, opts.workers),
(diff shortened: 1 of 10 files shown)


lib/supavisor/client_handler.ex
```diff
@@ -341,41 +342,21 @@ defmodule Supavisor.ClientHandler do
...
-        data.client_ready ->
-          {:next_state, :idle, data, handle_actions(data)}
-
-        opts.ps == [] ->
-          {:keep_state, data, {:timeout, 1_000, :wait_ps}}
-
-        true ->
-          {:keep_state, data, {:next_event, :internal, {:greetings, opts.ps}}}
-      end
+         {:ok, opts} <- Supavisor.subscribe(data.id) do
```
v0idpwn, line 345: We can't subscribe at this stage, because subscribing takes a slot in the pool. This would allow a slot to be taken before the authentication is performed, so an ill intentioned user can take slots from a pool owned by someone else, without knowing their password.

v0idpwn: At this stage, we can check for free slots, even wait for free slots to be available, but not take them

PR #1201, opened 2026-09-18
feat: startup options validation & conversion

### LINKED ISSUE #1198: Add an options schema, validate and convert options against it
Currently, all options are handled as strings. We must convert them according to https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-NAMES-VALUES.

For example, `-c jit=1` should be equivalent to `-c jit=true`.

Invalid options should be ignored and a NOTICE response should be sent to the client.

Snehil-Shah:
Adds startup options validation and type conversion for Supavisor-specific startup options to match how Postgres handles types in startup options: https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-NAMES-VALUES

Closes #1198


Snehil-Shah:
@v0idpwn One deviation from the original issue that I deliberately went with is killing a connection with a FATAL response instead of ignoring it with a NOTICE on invalid options. This is on par with how Postgres itself handles invalid options, and it also sits well with the current code semantics and how the startup packet was handled/rejected before this PR. WDYT?

test/supavisor/client_handler/protocol_helpers_test.exs
```diff
@@ -73,4 +96,10 @@ defmodule Supavisor.ClientHandler.ProtocolHelpersTest do
       end
     end
   end
+
+  # Builds a StartupMessage wire packet from key/value pairs.
+  defp startup_packet(pairs) do
+    body = Enum.map_join(pairs, fn {k, v} -> k <> <<0>> <> v <> <<0>> end)
+    <<0::32, 0, 3, 0, 0>> <> body <> <<0>>
+  end
```
v0idpwn, line 104: A bit better, IMO:
```suggestion
  defp startup_packet(pairs) do
    body = Enum.map(pairs, fn {k, v} -> [k, <<0>>, v, <<0>>] end)
    IO.iodata_to_binary([<<0::32, 0, 3, 0, 0>>,  body, <<0>>]
  end
```

v0idpwn:
@Snehil-Shah I'm worried that returning a FATAL may be a breaking change for users who are currently sending something invalid without issues. I understand Postgres does reject it, but since we were more lenient before, I don't think we should suddenly make it stricter.

Snehil-Shah:
@v0idpwn Makes sense, will switch to NOTICE. From the code's POV, FATAL did feel like the natural fit 😅, since sending a NOTICE needs a bit of extra plumbing to propagate the invalid options to the point where we can send the response back.

Snehil-Shah:
Example output with invalid options:

```
$ PGOPTIONS="-c jit=maybe -c log_level=trace" \
    psql "postgresql://postgres.dev_tenant@localhost:6543/postgres"
NOTICE:  invalid value for parameter "log_level": "trace"
HINT:  Available values: debug, info, notice, warning, error.
NOTICE:  parameter "jit" requires a Boolean value
psql (16.15)
Type "help" for help.

postgres=#
```

v0idpwn (approved):

v0idpwn:
Looks great! Merging.

v0idpwn:
@Snehil-Shah if you are interested in micro-optimizations, I think it would be a bit faster to pre-compute the booleans into a "truth table" map at compile time. The hot path then becomes a single map lookup `Map.fetch(@boolean_truth_table, val)`

If you do that please make sure to use Benchee to compare the options :) 

I'm not sure how much difference would it make.

Snehil-Shah:
@v0idpwn Interesting.. Trying it now, will report with the numbers :). Although I'm not sure how much it would benefit, as it just runs once during startup per connection for a few options and is not part of the hot path.

Snehil-Shah:
@v0idpwn It is pretty much the same. I ran the benchmark 5 times, and usually our current implementation **won**, although even truth table won a few runs. Pasting output of both cases below:

```
Name            ips        average  deviation         median         99th %
cond         889.94        1.12 ms     ±2.13%        1.11 ms        1.21 ms
table        841.07        1.19 ms     ±3.05%        1.18 ms        1.34 ms

Comparison: 
cond         889.94
table        841.07 - 1.06x slower +0.0653 ms
```

```
Name            ips        average  deviation         median         99th %
table        817.13        1.22 ms     ±5.27%        1.21 ms        1.47 ms
cond         798.92        1.25 ms    ±38.01%        1.19 ms        1.60 ms

Comparison: 
table        817.13
cond         798.92 - 1.02x slower +0.0279 ms
```

Curious about your thoughts..

Pasting the bench script below (correct me if there's anything wrong with my methodology):

```elixir
defmodule BoolCast do
  # current impl
  def cond_cast(value) do
    down = String.downcase(value)

    cond do
      down != "" and String.starts_with?("true", down) -> {:ok, true}
      down != "" and String.starts_with?("false", down) -> {:ok, false}
      down != "" and String.starts_with?("yes", down) -> {:ok, true}
      down != "" and String.starts_with?("no", down) -> {:ok, false}
      byte_size(down) >= 2 and String.starts_with?("on", down) -> {:ok, true}
      byte_size(down) >= 2 and String.starts_with?("off", down) -> {:ok, false}
      down == "1" -> {:ok, true}
      down == "0" -> {:ok, false}
      true -> :error
    end
  end

  # truth table
  @truth_table %{
    "t" => true, "tr" => true, "tru" => true, "true" => true,
    "y" => true, "ye" => true, "yes" => true,
    "on" => true, "1" => true,
    "f" => false, "fa" => false, "fal" => false, "fals" => false, "false" => false,
    "n" => false, "no" => false,
    "of" => false, "off" => false, "0" => false
  }
  def table_cast(value), do: Map.fetch(@truth_table, String.downcase(value))
end

values = ~w(1 t true on of off yes no maybe TRUE 0 f false)
workload = Stream.cycle(values) |> Enum.take(10_000)

Benchee.run(%{
  "cond" => fn -> Enum.each(workload, &BoolCast.cond_cast/1) end,
  "table" => fn -> Enum.each(workload, &BoolCast.table_cast/1) end
})
```

v0idpwn:
Thanks for trying it!

PR #1203, opened 2026-09-21
fix: stop accepting new pools when node starts shutting down

v0idpwn:
Nodes join an :accepting_pools syn scope on boot, held by NodeMembership. It is the last child of the application supervisor, so it terminates first on shutdown.

determine_node/2 now picks from the accepting nodes instead of every connected node, so a draining node stops receiving new pools.


lib/supavisor/node_membership.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  @impl true
+  def init(:ok) do
+    availability_zone = Application.get_env(:supavisor, :availability_zone)
+
+    :ok = :syn.join(:accepting_pools, :nodes, self(), node: node())
+    :ok = :syn.join(:availability_zone, availability_zone, self(), node: node())
+
+    {:ok, nil}
+  end
```
mentels, line 30: I'd add a `terminate/2` callback to emit a warning for a node leaving the scope. That would ease debugging.

mentels (approved):
Looks great!

Did you try some tests?

Do we expect some conflicts? After giving it a quick thought I gather the answer is _No_ — we don't expect duplicate node names from different nodes; am I right here?

lib/supavisor/node_membership.ex
v0idpwn: good call, will do

PR #1204, opened 2026-09-21
fix: drain pools before listeners

v0idpwn:
Ranch supervises the ClientHandlers, so stopping the listeners kills them.
The listeners are terminated before the pools, so by the time a pool's
Terminator ran its clients were already gone and there was nothing to drain.

Drainer sits just before NodeMembership in the application supervisor. It
drains every local pool in terminate/2 and blocks until they are done, so
clients finish their queries before the listeners go away.


lib/supavisor/drainer.ex
```diff
@@ -0,0 +1,62 @@
...
+
+    managers
+    |> Task.async_stream(&Manager.graceful_shutdown(&1, @drain_timeout),
+      max_concurrency: @max_concurrency,
+      timeout: :infinity,
+      on_timeout: :kill_task
+    )
+    |> Stream.run()
+
+    Logger.info("Finished draining pools")
```
mentels: I'd match on `{:exit, :timeout}` from the stream result and do `Logger.warning(... X failed to drain` so that we have visibility if it doesn't work smoothly

mentels (approved):
LGTM, one o11y nit

PR #1207, opened 2026-09-23
fix: rework protocol handling to better handle pipelining

v0idpwn:
In transaction mode, the DbHandler decided a batch was done by counting ReadyForQuery messages. The ClientHandler told it how many to expect (Q/S/F per write). Then, in #1205 we added an open_batch? flag for extended batches still waiting on a Sync. 

Unfortunately, that model breaks in several cases:
- Errors in extended batches: the backend skips everything up to the next Sync, so later Q messages in the same pipeline never produce the ReadyForQuery we counted on.
- COPY FROM STDIN (extended): the backend ignores Syncs sent during copy-in, so we expected ReadyForQuery messages that wouldn't come. LibPQ for example sends these Sync messages.
- Prepared statement messages weren't necessarily injected at the right points when using pipelining
- Races between writes: the DbHandler checked itself back into the pool as soon as it saw the final ReadyForQuery. A write the ClientHandler had already forwarded could then land on a connection that was back in the pool.

Additionally, a separate bug:
- Idle Sync short-circuit: the ClientHandler answered any packet starting with a Sync while idle, including a Sync pipelined together with a query.

On this PR, we follow the backend message by message instead of counting ReadyForQuery.
- FrontendMessageHandler records the tag of every relevant message forwarded in a write. Prepared statement packets are recorded as :ps placeholders, because the DbHandler decides what gets sent for them.
- Before forwarding, the ClientHandler sends DbHandler.expect_messages(pid, write_seq, tags). This replaces expect_ready_for_query/3.
- BackendMessageHandler keeps a queue of pending frontend messages. Each backend response consumes the,.

The ClientHandler now drives the release, again...

### Tests
The new integration tests in transaction_pipelining_test.exs cover:
 - Pipelined batches: batches that start with a Sync, bare Syncs, several Executes under one Sync, mixed simple and extended messages in one write, one message  per write, and writes split into small chunks.
- Errors: an error that doesn't affect the next batch, holding the backend until the Sync (including across a Flush), and a Query discarded after an extended error.
- Transactions spanning writes, in simple and extended protocol, including a failed transaction held until ROLLBACK.
- COPY FROM STDIN (simple and extended, with libpq's Syncs), CopyFail, a COPY that fails to start, a COPY that fails on bad data, and COPY TO STDOUT.
- Named prepared statements: close and re-parse in one write, reuse across writes and backends, and a Parse the backend already has, answered in its place in the pipeline.
- Races: the next write arriving before the previous reply, and an empty write racing the reply.


Snehil-Shah:
@v0idpwn okay, this is some eye-opening stuff. Love this!

